PR #12, opened 2026-05-01
fix: route slug-collision-masqueraded-as-error through existing retry

davesnowdon:
Lithos returns slug collisions as `status="error"` with a `Slug '...' already in use` message instead of the documented `status="slug_collision"`.  The existing FR-MCP-7 / AC-05-D slug-collision retry path was therefore never engaged, so the same articles re-failed every scheduled run with no progress.

Detect the slug-clash shape inside the error envelope and re-route to `slug_collision` so `_retry_slug_collision` re-runs with the disambiguating title suffix.  Logged separately so operators can spot the upstream classification bug and remove this shim once Lithos emits the correct status.

Refs staging incident 2026-05-01.

### DIFF
--- a/src/influx/lithos_client.py
+++ b/src/influx/lithos_client.py
@@ -73,6 +73,23 @@ def _first_non_empty_str(body: dict[str, Any], keys: tuple[str, ...]) -> str:
     return ""
 
 
+# Lithos currently bundles slug collisions into ``status="error"`` with a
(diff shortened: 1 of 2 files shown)


davesnowdon:
Not needed. Changed in Lithos instead https://github.com/agent-lore/lithos/pull/217

PR #84, opened 2026-05-04
fix: phrase-quote LCMA retrieve query disjuncts to neutralise Tantivy syntax

### LINKED ISSUE #80: Tantivy query parser fails on paper titles containing colons (silent search/dedup failure)
## Problem

In a 5-minute window at 06:05–06:06 on 2026-05-04, lithos-staging logged 23 "Tantivy query parse/search failed" warnings. Every failure has the same shape: a paper title containing a `Word: Subtitle` segment is fed unescaped to Tantivy, which interprets the capitalized word as a field qualifier:

```
WARNING Tantivy query parse/search failed: Field does not exist: 'Steps' | query='(When LLMs Stop Following Steps: A Diagnostic Study … | …)'
WARNING Tantivy query parse/search failed: Field does not exist: 'Memory' | query='(Persistent Visual Memory: Sustaining Perception for Deep Generation in LVLMs | …)'
WARNING Tantivy query parse/search failed: Field does not exist: 'RunAgent' | query='(RunAgent: Interpreting Natural-Language Plans with Constraint-Guided Execution | …)'
```

Plus 4 generic `Syntax Error` cases on titles containing single quotes / parentheses.

Affected fields the parser invented from titles in this single window: `Steps`, `Memory`, `RunAgent`, `GeoContra`, `Position`, `Harness`, `Memorize`, `Benchmarks`, `Yourself`, `BlenderRAG`, `MSACT`, `SAGA`, `Liberates`, `Images`, `Enough`, `Experts`, `Retrieval`.

## Why it matters

Academic titles routinely use the `Topic: Subtitle` form, so this silently breaks dedup/search lookups for a meaningful fraction of arXiv ingest. The query is rejected outright; nothing falls back to a substring match. Every failure is a silent data-quality miss for downstream dedup / similarity work.

The 06:05–06:06 burst lines up precisely with the staging-ai sweep that completed at 06:06:04 (`run_id=dfdc6b8a-…`, 32 ingested), so the query builder is almost certainly something on the influx side feeding raw paper titles into the lithos search API.

## Proposed fix

Escape Tantivy reserved characters (`:`, `(`, `)`, `"`, `'`, `\`) on whichever side constructs the query. The simplest and safest pattern is to either:

1. **Wrap each disjunct in double-quotes** so Tantivy treats it as a phrase query (and escape any `"` and `\\` inside), or
2. **Backslash-escape Tantivy reserved characters** before composing the `(a | b | c)` query.

Option 1 is easier and matches the intent (we want phrase-level matching against paper titles + claims, not field-qualified term searches).

## Where to look

- Whichever code path constructs the `(Title | Claim | Claim | …)` query — likely `src/influx/dedup.py` (we use Tantivy for dedup) or wherever lithos's search API is called from influx.
- If the escaping must live on the lithos side, file it there instead — but the influx caller should still defensively escape because it's the source of truth for the query inputs.

## Acceptance criteria

- A paper title like `"When LLMs Stop Following Steps: A Diagnostic Study"` does not produce a `Field does not exist` warning when fed to dedup/search.
- Test coverage for: titles with `:`, titles with single quotes, titles with parentheses, titles with double quotes (regression set drawn from today's failure list).
- The 23/day Tantivy parse warning rate drops to ~0 on staging.

## Refs

- 2026-05-04 06:05–06:06 lithos-staging Tantivy parse failure burst (23 occurrences in 5 minutes).

davesnowdon:
## Summary

- `compose_retrieve_query()` in `src/influx/lcma.py` previously joined the title and contributions with raw `" | "` separators. Academic titles using the `Topic: Subtitle` form caused Tantivy to interpret the capitalised word before the colon as a field qualifier and reject the query with `Field does not exist: '<Word>'`, silently zeroing the LCMA lexical scout result.
- This was the source of the 23 `Tantivy query parse/search failed` warnings observed in the 5-minute window at 06:05–06:06 on 2026-05-04 staging — see #80.

## Fix

- Wrap each disjunct (title + each contribution) as a Tantivy phrase query: double-quoted, with `\` and `"` backslash-escaped inside.
- Inside a phrase, `:`, `(`, `)`, `'` lose their syntactic meaning, so titles like `"RunAgent: Interpreting Natural-Language Plans"` parse cleanly.
- The `|` between disjuncts is still parsed as OR, so the caller's intent (OR of phrase matches across title + first 3 contributions) is preserved.

## Scope

Caller-side fix only, per the investigation comment on #80. The server-side hardening for `lithos_retrieve` / the LCMA lexical scout is tracked separately at agent-lore/lithos#240.

## Test plan

- [x] `make check` — lint + format + pyright clean, **1733 tests pass**
- [x] Existing 13 `GOLDEN_CASES` rows updated to expect the new wrapped form (behaviour-change to the wire format, which is the contract being changed)
- [x] 7 new `ISSUE-80-*` rows cover the regression set called out in the issue's acceptance criteria:
  - colon in title (`Topic: Subtitle`)
  - colon in title with contributions
  - single quote/apostrophe in title
  - parentheses in title
  - double quotes inside content (escaped)
  - backslashes inside content (escaped)
  - colon in contribution (not just title)
- [x] `TestTruncation` updated for the new wrapped lengths — the truncation contract remains a simple slice of the final composed string with no word re-wrap

Closes #80

### DIFF
--- a/src/influx/lcma.py
+++ b/src/influx/lcma.py
@@ -31,6 +31,18 @@ _WHITESPACE_RE = re.compile(r"\s+")
 _ARXIV_RE = re.compile(r"arXiv:(\d{4}\.\d{4,5}(?:v\d+)?)")
 
 
+def _phrase_quote(s: str) -> str:
+    """Wrap *s* as a Tantivy phrase query, backslash-escaping ``\\`` and ``"``.
+
+    Phrase quoting neutralises Tantivy reserved characters inside the
+    payload (notably ``:`` from ``Topic: Subtitle`` titles, plus ``(``,
+    ``)``, ``'``), so the parser cannot mistake content for field
+    qualifiers or syntax (#80).
+    """
+    escaped = s.replace("\\", "\\\\").replace('"', '\\"')
+    return f'"{escaped}"'
+
+
 def compose_retrieve_query(
     title: str,
     contributions: list[str] | None = None,
@@ -42,9 +54,13 @@ def compose_retrieve_query(
        trim each, and skip any that are empty after trimming. Empty
        elements within those first three are dropped, NOT replaced by
        later non-empty entries (FR-LCMA-2 step 2).
-    3. Join the surviving parts with ``" | "``.
-    4. Collapse internal whitespace runs to a single space.
-    5. Truncate to 500 characters (simple slice, no word re-wrap).
+    3. Collapse internal whitespace runs to a single space within each
+       disjunct.
+    4. Wrap each disjunct as a Tantivy phrase (double-quoted, with
+       ``\\`` and ``"`` escaped) so reserved characters in titles like
+       ``Topic: Subtitle`` cannot be parsed as field qualifiers (#80).
+    5. Join the wrapped parts with ``" | "``.
+    6. Truncate to 500 characters (simple slice, no word re-wrap).
     """
     parts: list[str] = [title]
 
@@ -55,8 +71,8 @@ def compose_retrieve_query(
                 continue
             parts.append(stripped)
 
-    composed = " | ".join(parts)
-    composed = _WHITESPACE_RE.sub(" ", composed)
+    quoted = [_phrase_quote(_WHITESPACE_RE.sub(" ", p).strip()) for p in parts]
+    composed = " | ".join(quoted)
     return composed[:_MAX_QUERY_LEN]
 
 
--- a/src/influx/lcma.py
+++ b/src/influx/lcma.py
@@ -31,6 +31,11 @@ _WHITESPACE_RE = re.compile(r"\s+")
 _ARXIV_RE = re.compile(r"arXiv:(\d{4}\.\d{4,5}(?:v\d+)?)")
 
 
+def _escape_phrase_inner(s: str) -> str:
+    """Backslash-escape ``\\`` and ``"`` for inclusion inside a phrase query."""
+    return s.replace("\\", "\\\\").replace('"', '\\"')
+
+
 def _phrase_quote(s: str) -> str:
     """Wrap *s* as a Tantivy phrase query, backslash-escaping ``\\`` and ``"``.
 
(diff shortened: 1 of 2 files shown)


davesnowdon:
Pushed a follow-up commit addressing the reviewer's truncation finding.

**Bug**: the original commit blindly sliced the joined query at 500 chars, which could split a phrase mid-string and leave an unmatched opening `"` — re-introducing the same parse-failure class the patch was meant to eliminate. The earlier `TestTruncation` tests asserted that malformed shape, baking the bug into the contract.

**Fix**: truncation now happens *before* wrapping, in two layers:

1. **Title is always emitted, fitted to budget.** A new `_fit_phrase(content, budget)` helper escapes the content, then if the wrapped length exceeds budget, shaves from the tail of the *escaped* form. It also tracks trailing backslashes — if the truncation would split a `\\` or `\"` escape (odd-count trailing run), it shaves one more to keep the escaped string valid.
2. **Each contribution is all-or-nothing.** A contribution is included only if its wrapped form plus separator fits whole in the remaining budget; otherwise it is dropped, never sliced.

**Tests rewritten** to assert balanced quotes instead of the malformed shape:

- `test_long_title_truncated_to_500_with_balanced_quotes` — 600-char title produces `'"' + "x"*498 + '"'` (matched quotes, len 500).
- `test_title_at_exact_budget_unchanged` — 498-char title wraps to exactly 500.
- `test_title_one_over_budget_loses_one_char` — 499-char title loses one inner char.
- `test_title_truncation_does_not_leave_dangling_backslash` — title of 300 backslashes (escapes to 600) shaves to even-count parity.
- `test_long_composed_drops_contribution_that_will_not_fit` — 200-char contribution after a 400-char title is dropped, not sliced.
- `test_partial_fit_keeps_fitting_contributions_drops_rest` — first contribution kept, second over-budget contribution dropped.
- `test_result_always_has_balanced_quotes` — parametric stress at boundary sizes (1, 100, 250, 497–501, 600, 1000).

`make check`: 1738 passed (5 new truncation tests on top of the prior 1733).

Verified manually:
```
>>> compose_retrieve_query("x"*499)
'"xxxx...xxxx"'  # len 500, both quotes intact
```

PR #133, opened 2026-05-10
fix: reject invalid RSS entry URLs before archive/extraction (#131)

### LINKED ISSUE #131: Reject invalid RSS entry URLs before archive/extraction acquisition
## What to build

Reject or quarantine malformed RSS entry article URLs before the RSS acquire path attempts archive download and article extraction.

Staging evidence after the OpenAI 429 fix showed accepted RSS items from the Sourcegraph blog carrying article links like `http://localhost:5174/...`. Influx accepted those items, then attempted archive download and extraction against the localhost URL, producing degraded `archive_acquisition` runs and ingested notes with `influx:archive-missing` instead of rejecting the bad upstream item.

This is happening because the RSS parser currently trusts `entry.link` verbatim in `parse_feed()` and `build_rss_note_item()` then uses that URL for both `download_archive()` and `extract_article()`:

- `src/influx/sources/rss.py:105-125`
- `src/influx/sources/rss.py:209-241`
- `src/influx/storage.py`
- `src/influx/extraction/article.py`
- `src/influx/urls.py`

The fix should add an explicit candidate/article URL validation policy for RSS entry links before expensive acquisition starts. At minimum, obviously broken or non-production article URLs such as loopback/private-host/self-dev URLs should not flow through the normal ingestion path.

The implementation should preserve the distinction between:

- a feed fetch succeeding
- an individual feed item being unusable because its article URL is invalid

so operators can tell whether the upstream feed itself was reachable but emitted bad item links.

## Acceptance criteria

- [ ] RSS item processing validates article URLs before `download_archive()` / `extract_article()` are attempted.
- [ ] Loopback, private-host, malformed, or otherwise disallowed RSS entry URLs are not ingested as ordinary accepted items with `influx:archive-missing`.
- [ ] Run / log diagnostics make it clear that the feed fetched successfully but specific items were rejected because their article URLs were invalid.
- [ ] Tests cover at least one malformed/loopback RSS entry URL and verify that the item is rejected before archive/extraction acquisition work starts.
- [ ] Existing valid public RSS item URLs still flow through the current acquire and ingest path unchanged.

## Blocked by

None - can start immediately.

davesnowdon:
## Summary

- New syntactic URL validator `classify_article_url()` in `src/influx/urls.py` rejects RSS entry links whose hostname is loopback / private / link-local / multicast, plus malformed URLs and disallowed schemes (anything outside http/https). Pure-syntactic — no DNS — using `ipaddress` stdlib classification labels matching `http_client._classify_ip()` for log/dashboard consistency.
- `parse_feed()` now calls the validator before each item enters the pipeline. Rejected items are dropped with a structured WARNING log, increment a per-run counter, and bump a new metric — they no longer trigger `download_archive()` / `extract_article()` and never produce an `influx:archive-missing` note.
- Honors the existing `allow_private_ips` config flag (mirrors the SSRF guard) so test fixtures using `FakeArticleServer` on `127.0.0.1` keep working.
- Run-ledger entry gets `invalid_url_rejections_total` so operators can tell "feed fetched OK, item links upstream-malformed" apart from a transient archive failure on an otherwise-valid URL.

## Why

Staging incident: the Sourcegraph RSS feed (after the OpenAI 429 fix) emitted article links like `http://localhost:5174/...`. Influx accepted those items, attempted archive download + article extraction against localhost, and produced degraded `archive_acquisition` runs with notes tagged `influx:archive-missing`. Validation only happened deep inside `download_archive()` via the SSRF guard — too late.

## Test plan

- [x] `uv run pytest` — 1838 passed.
- [x] `uv run pyright` on touched files — 0 errors.
- [x] `uv run ruff check` + `ruff format --check` on touched files — clean.
- [x] New unit cases in `tests/unit/test_url_normalisation.py` for `classify_article_url` (public passes; loopback / private / link-local / multicast / scheme / malformed rejected; `allow_private` opt-in).
- [x] New unit cases in `tests/unit/test_rss_fetcher.py` for `parse_feed` (mixed-feed XML drops localhost + RFC1918 items; `allow_private_ips=True` keeps them; counter increments inside a run context; safe outside one).
- [x] Existing `tests/integration/test_rss_to_lithos.py` updated to pass `allow_private_ips=True` to `parse_feed` (it already runs with `SecurityConfig(allow_private_ips=True)`); all three integration cases pass.

## Notes for review

- `archive-missing` tagging path is intentionally untouched; it remains the safety net for transient archive failures on valid URLs.
- DNS-based rejection of hostnames that resolve to private addresses stays the SSRF guard's job at fetch time (defence in depth) — the new check is fast and synchronous so `parse_feed` stays cheap.
- New counter is initialised in `run_service.ledger_lifecycle` next to `current_filter_errors` and `current_fetched_total`; threaded through `RunLedger.complete` / `_finish` and the active+finalised entry shape. `ledger.fail` / `ledger.skip` paths leave the field as `None`, mirroring `archive_failures_total`.

Closes #131


davesnowdon:
Addressed both review concerns in `e453798`:

**Concern 1 (High) — staging `allow_private_ips=true` defeating the fix.**
Removed the production thread-through: `_fetch_rss_feed` no longer accepts `allow_private_ips`, and the provider closure no longer passes `config.security.allow_private_ips` into `parse_feed`. The two flags have different jobs — `allow_private_ips` is the fetcher's escape hatch for *configured* internal services, not "trust upstream feeds to advertise localhost links." `parse_feed`'s `allow_private_ips` parameter stays only as a test-direct opt-in for the `FakeArticleServer` integration tests. New regression test `TestProductionPathIgnoresAllowPrivateIps` exercises `make_rss_item_provider` end-to-end with `SecurityConfig(allow_private_ips=True)` plus a feed body containing the Sourcegraph-style `http://localhost:5174/leak` link and asserts that only the public item reaches the build step.

**Concern 2 (Medium) — `fetched_total` undercount + `fetch_stall` misclassification.**
- `parse_feed` now also calls `record_fetched_items(1)` per rejection, so the run's pre-filter `fetched_total` reflects what the feed actually returned, not just what survived URL validation. A feed that returned 5 items with every URL bad now records `fetched_total=5, invalid_url_rejections_total=5`, not `fetched_total=0`.
- New degraded reason **`invalid_url_stall`** in `RunLedger.complete()`. Fires (single-run signal, no consecutive-runs ratchet) when `fetched_total > 0 AND items_reached_filter == 0 AND invalid_url_rejections_total > 0`. Mutually exclusive with `filter_stall` (the existing branch is now also gated on `not has_invalid_url_stall`) and with `fetch_stall` (which still requires `fetched_total == 0`). `run_service` emits a corresponding WARNING and metric increment.
- New ledger tests cover: all-URL-rejected fires `invalid_url_stall`; partial rejection still fires `filter_stall` because items reached the filter; `fetched_total == 0` still fires `fetch_stall`; first-run firing (no historical ratchet).

Verification: `uv run pytest` 1845 passed, `uv run pyright` 0 errors, ruff lint+format clean.

PR #136, opened 2026-05-10
fix: route RSS notes through repair sweep with source-aware hooks

### LINKED ISSUE #130: Give degraded RSS notes a real repair path instead of unsupported-source sweep retries
## What to build

Extend the repair sweep so RSS notes degraded by archive/text acquisition can be repaired, or explicitly terminalized, instead of re-entering the sweep indefinitely with unsupported-source behavior.

Recent staging runs are still degrading on `archive_acquisition`, and many of those degraded items are RSS notes. The current repair implementation is largely arXiv-only:

- `archive_download` retry explicitly requires arXiv source in `src/influx/repair_hooks.py:302-345`
- `text_extraction` retry explicitly requires arXiv source in `src/influx/repair_hooks.py:348-380`

That matches the staging warnings seen during repair sweeps (`source not supported`, `No archive path found`, `No ## Full Text section found`). In practice this means many degraded RSS notes have no real recovery path today, even though runs continue and ingest them.

The goal of this issue is to make RSS degradation operationally honest:

- either RSS archive/text failures become repairable via source-appropriate hooks
- or the sweep terminates / classifies them cleanly so they do not produce repeated noisy retries with no chance of success

The implementation should preserve the existing arXiv repair behavior.

## Acceptance criteria

- [ ] RSS notes tagged for archive/text repair no longer loop indefinitely through unsupported-source repair paths.
- [ ] There is a defined RSS behavior for degraded archive/text notes: successful retry, explicit terminalization, or another clearly documented end state.
- [ ] Repair logs and counters make RSS repair outcomes understandable to operators.
- [ ] Tests cover at least one degraded RSS note shape and verify the new repair outcome.
- [ ] Existing arXiv repair behavior remains intact.

## Blocked by

None - can start immediately.

davesnowdon:
Closes #130.

## Summary

- Replace the arxiv-only gate in `_make_archive_download_hook` and `_make_text_extraction_hook` with a per-source dispatcher so RSS notes go through the same retry / cap / terminal-flip machinery as arxiv.
- RSS resolver derives `download_archive` kwargs from state already on disk: `source_url` from YAML frontmatter, item_id from the note's `rss-<feed>-<hash>` id (mirroring arxiv's `arxiv-<id>` convention), year/month from the note path. `_NOTE_PATH_RE` now matches both `papers/` and `articles/` prefixes.
- RSS text extraction reuses `extract_article` (the same helper RSS uses at acquisition) and returns `text:html` so the note exits the text-extraction stage.
- Failure classification uses the existing source-agnostic `_classify_download_kind`: `oversize`/`parse`/`validate` count toward the 3-attempt cap and flip `influx:archive-terminal`; HTTP and network kinds stay transient and re-enter the sweep next pass — identical policy to arxiv.

## Why

Staging runs were degrading on `archive_acquisition` and many of those degraded notes were RSS. The arxiv-only gate raised `unsupported_source` every pass; the text path terminalised cleanly via `_terminate_unsupported_text_source`, but the archive path treated `unsupported_source` as transient (it isn't in `_COUNTED_STAGES`), so `influx:archive-missing` re-selected the note forever. The fix replaces the gate with a dispatcher and lets RSS flow through the existing classify-and-cap pipeline.

## Acceptance criteria

- [x] RSS notes tagged for archive/text repair no longer loop through unsupported-source paths — they hit real retries and the existing counted-stage cap.
- [x] Defined RSS behavior: `success` (archive_path written, `archive-missing` cleared), `counted-cap-terminal` (3× oversize/parse/validate → `influx:archive-terminal`), `transient-retry` (HTTP/timeout/network → re-enter next pass), or `resolve` (missing source_url / id prefix / path year-month → operator hand-fix).
- [x] Repair logs and counters now reflect real RSS failure stages (`stage=http`, `stage=timeout`, `stage=oversize`, …) via the existing `_log_stage_failure` and `RepairCounters` instead of a single `unsupported_source` repeating forever. No metric schema change.
- [x] Tests cover at least one degraded RSS note shape — see new `TestArchiveDownloadHookRss*` and `TestTextExtractionHookRss*` classes (10 new test cases).
- [x] Existing arxiv repair behavior intact — arxiv hook path is byte-equivalent; all prior tests pass unchanged.

## Test plan

- [x] `uv run pytest tests/unit/test_repair_hooks_production.py -v` — 55 passed (10 new RSS tests)
- [x] `uv run pytest tests/unit/test_repair_sweep.py tests/unit/test_repair_stage_selection.py tests/unit/test_repair_counters.py tests/unit/test_repair_hooks.py` — 113 passed
- [x] `uv run pytest` (full suite) — 1875 passed
- [x] `uv run ruff check src/influx/repair_hooks.py tests/unit/test_repair_hooks_production.py` — clean
- [ ] Staging: grep `sweep_stage=archive_download` filtered to `source:rss-*` notes; expect a mix of real stages instead of `unsupported_source` looping

### DIFF
--- a/src/influx/repair_hooks.py
+++ b/src/influx/repair_hooks.py
@@ -22,6 +22,7 @@ import trafilatura
 from influx.config import AppConfig
 from influx.enrich import tier3_extract as _tier3_extract
 from influx.errors import ExtractionError, LCMAError, NetworkError
+from influx.extraction.article import extract_article
 from influx.extraction.html import _clean_html_fragments, _strip_tags
 from influx.extraction.pdf import extract_pdf
 from influx.extraction.pipeline import extract_arxiv_text
@@ -193,7 +194,11 @@ def _extract_from_archive(
 
 _ARXIV_ID_TAG_PREFIX = "arxiv-id:"
 _SOURCE_TAG_PREFIX = "source:"
-_NOTE_PATH_RE = re.compile(r"papers/(?P<source>[^/]+)/(?P<year>\d{4})/(?P<month>\d{2})")
+_NOTE_PATH_RE = re.compile(
+    r"(?:papers|articles)/(?P<source>[^/]+)/(?P<year>\d{4})/(?P<month>\d{2})"
+)
+_RSS_NOTE_ID_PREFIX = "rss-"
+_SOURCE_URL_FRONTMATTER_KEY = "source_url:"
 
 
 def _find_tag(tags: list[str], prefix: str) -> str | None:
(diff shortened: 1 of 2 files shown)


davesnowdon:
Good catch — the reviewer is right that the `TextExtractionHook` protocol (`repair.py:283-285`) explicitly contracts a `text:abstract-only` return on cascade fall-through, and the previous RSS implementation violated that by propagating `ExtractionError`, which the sweep treats as transient → infinite loop on a permanently broken live page.

Fixed in c340988:

- On any `ExtractionError`/`NetworkError` from `extract_article(...)`, the RSS hook now logs a structured WARN with the failure stage/kind and returns `"text:abstract-only"`. The sweep stamps the tag, the note exits the text-extraction stage, and the loop converges.
- Operator visibility preserved via the WARN log (covered by a new `test_logs_warning_on_failure` test).
- Recovery path is intact: once `archive_download` lands an HTML archive on disk, the source-agnostic `re_extract_archive` hook (via the `abstract_only_reextraction` stage) upgrades the tag to `text:html`. If archive is permanently unrecoverable, `text:abstract-only` is the appropriate end state.
- `resolve`-stage failures (missing `source_url` frontmatter) still raise — those are operator-fix-required structural issues, not extraction failures, and should keep surfacing through `_log_stage_failure`.

The two prior failure tests now assert the convergence contract instead of the propagation behaviour:
- `test_extraction_error_returns_abstract_only`
- `test_network_error_returns_abstract_only`

Worth noting: the existing arxiv `text_extraction` hook has the same theoretical loop on `extract_arxiv_text` cascade fall-through. That's a pre-existing issue out of scope for #130 — happy to file a separate one if useful.

PR #180, opened 2026-05-24
feat: rewrite-legacy-notes diagnose subcommand (#176)

davesnowdon:
## Summary

Closes #176.

Adds `./scripts/influx-diagnose.py rewrite-legacy-notes` to repair the ~170 pre-fix Influx notes (created 2026-05-05 → 2026-05-07) whose outer Lithos doc-level metadata is empty (`tags=[]`, `source_url=null`, `confidence=0.0`) but whose `content` carries an embedded YAML frontmatter block with the correct values. Without the fix, the staging slug-collision backlog (1,459 entries and growing) doesn't drain because `_classify_squatter` (lithos_client.py:244) can't match these zombie squatters by `source_url` or `arxiv-id:` and falls through to `distinct` → suffix retry → backlog append.

The fix is data-only — the writer bug itself has been resolved since 2026-05-08, but the residue docs remain on disk. After this lands and runs with `--apply`, the recovery chain classifies each rewritten doc as `duplicate` via source_url equality, the backlog stops growing, and the promotion-gate `unexpected_failure_present` reason clears as old failed runs roll off the gate window.

## What's in the PR

- **`scripts/influx-diagnose.py`** — new `cmd_rewrite_legacy_notes` subcommand mirroring `cmd_squatters`'s safety contract (dry-run by default, `--apply` requires `--yes <doc-id>` / `--yes-to-all` / `--id`). Uses `client.call_tool("lithos_write", args)` directly for UPDATE-by-id (matching the `influx.repair::_sweep_call_write` pattern; `LithosClient.write_note` doesn't accept `id`/`expected_version`).
- **Helpers** — `parse_legacy_note_frontmatter` (reuses `influx.notes._split_frontmatter` + `yaml.safe_load`), `strip_legacy_frontmatter` (drops embedded frontmatter AND inner `# Title` line so Lithos's title-prepend doesn't produce double headings), `_select_legacy_doc_ids_from_backlog` (parses `unresolved-slug-collisions.jsonl` for both `first_existing_id` and `retry_existing_id`).
- **Outcome buckets** — `would_rewrite` / `rewrote` / `skipped_already_fixed` (idempotency guard) / `skipped_unparseable` / `refused_non_influx_authored` / `partner_owns_url` (the colliding partner doc already landed this source_url — expected during batch apply, not a failure) / `failed`.
[...]

Issue #179, opened 2026-05-24
Pin the 2026-05-07/08 fix commit and add a write-call-shape regression test


davesnowdon:
## Background

Spun out of #176. During the #176 audit we confirmed an Influx writer-side bug existed between **2026-05-05 and 2026-05-07** that produced notes with empty doc-level `tags` / `source_url` / `confidence` (~245 broken docs on disk). The bug was resolved between 2026-05-07 and 2026-05-08, but the exact fix commit was not pinned — the diagnosis came from comparing on-disk corpus state across the window, not from reading the fix.

Empty-tag rate by `created_at` date (from a corpus survey of `/home/dns/projects/lithos/data/staging/lithos/knowledge/articles/`):

```
2026-05-05  14% empty  (75 / 527)
2026-05-06  32% empty   (9 /  28)
2026-05-07  40% empty (161 / 402)
2026-05-08   0% empty  (34 /  34)   ← fix landed
2026-05-09 .. 2026-05-24:  0% empty across 533 new docs
```

We need to (a) pin which commit fixed this and (b) add a regression test that locks in the contract so the original shape cannot silently re-emerge.

## Investigation — pin the fix commit

Candidates between 2026-05-07 and 2026-05-08:

```
50a205a 2026-05-08 restart policy (#123)
6b3adca 2026-05-08 refactor: deduplicate repair sweep stage execution (#122)
d2f03bd 2026-05-08 Back off and retry on filter model 429s (#121)
c854f18 2026-05-08 Improve diagnostics for webhook failures and LCMA skips (#120)
3139b30 2026-05-08 Centralize Lithos MCP response decoding (#118)
39f6f4f 2026-05-08 Flag archive acquisition failures in run outcomes (#117)
735f452 2026-05-08 Terminalize unsupported repair sources (#119)
b1856d3 2026-05-07 Harden final Lithos task completion handling (#115)
bd1b299 2026-05-07 Fix LCMA edge_upsert MCP argument shape (#116)
f03998b 2026-05-07 Fix unresolved slug collision backlog persistence (#114)
```

Most likely candidates: **`3139b30`** (Centralize Lithos MCP response decoding) or **`6b3adca`** (refactor: deduplicate repair sweep stage execution). Both touch the write/decoding paths in a way that could resolve a doc-level-metadata-drop bug.

Suggested approach:

1. Diff each candidate against its parent — specifically look at any change in how `write_note`, `_retry_*`, or sweep-rewrite args are constructed.
[...]

Issue #178, opened 2026-05-24
Renderer: emit body-only content and reject embedded frontmatter at the lithos_write boundary


davesnowdon:
## Background

Spun out of #176. Influx's renderer (`src/influx/renderer.py:249`) builds note content as `f"---\n{frontmatter}\n---\n# {title}\n..."` and passes the entire string to `lithos_write` as the `content` argument. Lithos spec §5.1 defines `content` as **"Markdown content (without frontmatter)"** — structured fields (`tags`, `source_url`, `confidence`, etc.) are first-class API parameters, and frontmatter is Lithos's to own.

Today's Lithos tolerates the embedded `---`-fenced block in `content` (the outer doc-level metadata still lands correctly from API parameters), so this is **not a bug** in current code. But it is a spec violation worth resolving:

- Wastes ~200 bytes/note on duplicated state that's never read back.
- Contaminates `lithos-enrich`'s entity extraction — pre-fix docs have `entities: [Profile, Summary, Score, User, Notes, ...]` populated from the embedded frontmatter's section headings.
- Forms a latent regression risk: if Lithos's frontmatter handling ever changes (stricter validation, different YAML parser, etc.), the original 2026-05-05→07 bug shape could re-emerge silently.

### 1. Renderer: body-only output

In `src/influx/renderer.py`:

- `render_note` (line 167) returns markdown **body only** — drop the `---\n{frontmatter}\n---\n` prefix at line 249. The `# {title}` heading and all `## ...` sections / archive / repair / user-notes regions stay as-is.
- `_render_frontmatter` (line 126) — two options to pick from at implementation time:
  - **Option A1 (split helper):** keep it exported under a name like `render_frontmatter_for_inspection` so tests and any future tooling can still produce the YAML shape, but production `render_note` no longer calls it.
  - **Option A2 (delete):** drop `_render_frontmatter` entirely; tests are rewritten to assert on body content; any tooling that needs the YAML shape constructs it ad-hoc.

Option A1 is the lower-friction landing; A2 is the cleaner end-state. Pick at implementation time.

### 2. Strict-mode assertion in `LithosClient.write_note`

In `src/influx/lithos_client.py`:

- `write_note` (line 767) — add a defensive check at the top: if `content.startswith("---\n")`, raise a clear error (e.g. `EmbeddedFrontmatterError` or similar). This is a contract assertion, not a defensive runtime guard — production code should never hit it once the renderer change lands.
[...]

Issue #176, opened 2026-05-24
Rewrite ~245 legacy notes with empty doc-level metadata to drain slug-collision backlog


davesnowdon:
## TL;DR

A writer bug existed in Influx between **2026-05-05** and **2026-05-07** that produced ~245 notes with empty doc-level `tags` / `source_url` / `confidence` (despite the rendered content carrying the correct values inside an embedded YAML frontmatter block). **The bug was fixed by 2026-05-08** and every Influx write since has landed clean metadata. The 1,459-entry slug-collision backlog accumulating today is healthy current writes colliding with those 245 historical zombie squatters — the slug-recovery chain can't classify them as duplicates because `_classify_squatter` matches on `source_url` and `arxiv-id:` tag, both of which are empty on the affected docs.

**The Influx-side fix is a one-shot rewrite job for the ~245 broken docs.** Once they have proper doc-level metadata, the next sweep's `_classify_squatter` will recognise each as `duplicate`, the slug collision auto-resolves, no backlog append. The backlog drains on the following sweep.

## Symptom

```
$ ./scripts/influx-diagnose.py --env staging promotion-gate
Promotion gate: FAIL
  reason          : unexpected_failure_present
  severity_counts : success=10 expected_lossy=4 unexpected_failure=6
  top reasons     : invalid_note_state=5 archive_acquisition=2 source_acquisition=2 ingestion_stall=1
```

`unresolved-slug-collisions.jsonl` has grown to **1,459 entries since 2026-05-08**. ~5 new entries land each day; today (2026-05-24) added 5.

## Evidence — bug window is bounded

Surveyed every Influx-authored doc on disk under `/home/dns/projects/lithos/data/staging/lithos/knowledge/articles/`, grouped by `created_at` date:

```
created_at  empty_tags / total  rate
2026-05-05   75 /  527         14%
2026-05-06    9 /   28         32%
2026-05-07  161 /  402         40%
2026-05-08    0 /   34          0%   ← fix landed
2026-05-09    0 /  150          0%
2026-05-10    0 /   37          0%
... through 2026-05-24:        0% across 533 new docs
```

The bug stopped reproducing between 2026-05-07 and 2026-05-08 — no commit on the writer path explicitly calls out a "missing tags" fix, so the resolution was incidental to one of:

```
$ git log --after="2026-05-07" --before="2026-05-09" --pretty="%h %ad %s" --date=short
50a205a 2026-05-08 restart policy (#123)
[...]

PR #184, opened 2026-05-24
feat: --corpus-scan selector for rewrite-legacy-notes (#181)

### LINKED ISSUE #181: Add --corpus-scan to rewrite-legacy-notes (selector beyond the backlog file)
## Background

Spun out of #176. The merged ``rewrite-legacy-notes`` subcommand selects candidate doc ids from ``${INFLUX_STATE_PATH}/unresolved-slug-collisions.jsonl`` (deduplicated). That selector is good for the **active blockers** — docs whose slugs have already caused collisions and landed in the backlog — but misses docs that:

- Are still broken (empty doc-level ``source_url``)
- Have intact embedded frontmatter (so they're recoverable)
- Just haven't collided yet

In staging tonight we found this gap to be substantial: 351 recoverable docs in the corpus vs 144 in the backlog selector. The other 207 sit in the on-disk corpus until the repair sweep or a colliding ingest surfaces them — at which point they become part of the slow trickle of new backlog entries (~5/run before the #180 fix, ~0–6/run after).

Tonight we worked around the gap by:

```bash
# Scan corpus for empty-source_url Influx-authored docs with parseable inner FM.
uv run python -c "..." > /tmp/recoverable-ids.txt   # 351 ids

# Pass them all via --id (works but ugly).
ID_FLAGS=$(awk '{printf " --id %s", $1}' /tmp/recoverable-ids.txt)
./scripts/influx-diagnose.py --env staging rewrite-legacy-notes --apply $ID_FLAGS
```

This should be a first-class operator workflow.

## Proposed change

Add ``--corpus-scan`` to ``cmd_rewrite_legacy_notes``. When set:

1. Resolve ``${LITHOS_KNOWLEDGE_PATH}`` (or read from the env file) and recurse the ``articles/`` subtree.
2. For each ``*.md`` parse the outer YAML frontmatter (the file *is* a frontmatter-prefixed markdown file Lithos owns; the format is stable).
3. Include the doc id when ``author == 'influx'`` AND ``source_url`` is empty.
4. Default to dry-run as today; ``--apply`` writes via the existing path.
5. Mutually exclusive with ``--id`` (corpus-scan ignores explicit ids; ``--id`` overrides the scan).

The actual rewrite path is unchanged — only the candidate selector changes.

## Tests

- ``test_corpus_scan_finds_broken_docs`` — fixture corpus with mixed influx/non-influx, empty/populated source_url; selector returns only the broken influx-authored ones.
- ``test_corpus_scan_respects_id_override`` — when both ``--id`` and ``--corpus-scan`` are passed, ``--id`` wins (or the script rejects the combination — pick one at impl time).

## Acceptance criteria

- [ ] ``./scripts/influx-diagnose.py --env staging rewrite-legacy-notes --corpus-scan`` enumerates every broken Influx-authored doc on disk, regardless of backlog presence.
- [ ] Dry-run is the default; ``--apply --yes-to-all`` rewrites the full set.
- [ ] Existing backlog-derived selector continues to work unchanged (default when ``--corpus-scan`` and ``--id`` are both absent).

## Cross-references

- #176 — original rewrite-legacy-notes work where this gap was discovered.
- #178 — renderer body-only output (unrelated to selector but in the same general cleanup track).
- #179 — pin the fix commit + regression test.

davesnowdon:
## Summary

Closes #181.

Adds a `--corpus-scan` flag to `./scripts/influx-diagnose.py rewrite-legacy-notes`. When set, the selector walks the on-disk Lithos `articles/` subtree and surfaces every Influx-authored doc with empty doc-level `source_url` — including recoverable docs that haven't collided yet.

The merged subcommand's default selector reads the slug-collision backlog file, which only catches the *actively blocking* docs. During staging cleanup we found this gap to be substantial: 351 recoverable docs in the corpus vs ~170 in the backlog selector. The workaround at the time was an inline corpus scan piped through `--id` flags — functional but unergonomic and tied to a specific shell incantation.

## What's in the PR

| File | Change |
|---|---|
| `scripts/influx-diagnose.py` | Two new helpers (`_resolve_corpus_articles_path`, `_select_legacy_doc_ids_from_corpus`); new `--corpus-scan` argparse flag; runtime mutual-exclusion guard with `--id`; updated empty-result messaging. |
| `tests/unit/test_diagnose_rewrite_legacy_notes.py` | 13 new tests across three classes (`TestSelectLegacyDocIdsFromCorpus`, `TestCorpusScanFlagWiring`, `TestResolveCorpusArticlesPath`). |
| `docs/operations/staging-runbook.md` | New §8f documenting the subcommand, its three selector modes, outcome buckets, and safety contract. |

## Selector resolution

| Selector | Behaviour |
|---|---|
| Default (no flag) | Reads `${INFLUX_STATE_PATH}/unresolved-slug-collisions.jsonl`. Triage shape. |
| `--id <doc-id>` (repeatable) | Targets known doc ids directly. |
| `--corpus-scan` | Walks `$LITHOS_KNOWLEDGE_PATH/articles` (or the staging-convention fallback at `$INFLUX_ARCHIVE_PATH/../lithos/knowledge/articles`). Broader cleanup shape. Mutually exclusive with `--id`. |

The actual rewrite path is unchanged — only the candidate selector changes.

## Smoke test against staging

```
$ LITHOS_URL=http://localhost:8766/sse \
  ./scripts/influx-diagnose.py --env staging rewrite-legacy-notes --corpus-scan
Dry-run mode: processing 1 doc id(s) via lithos_read (no writes) on http://localhost:8766/sse

                 would_rewrite  doc_id=867ecc1b-9b00-4b47-9720-a21822b7de3e  ready for rewrite  tags=7  source_url=https://neo4j.com/developer-blog/claude-converses-neo4j-via-mcp  confidence=1.0  content_bytes=53762→53424

Summary: would_rewrite=1
```

That one residual doc (Neo4j blog post, created 2026-05-07) slipped through the earlier corpus-scan-via-bash workaround because the repair sweep had updated it between the scan and the apply. The new flag naturally re-surfaces missed docs on every run.

## Test plan

- [x] `uv run pytest tests/unit/test_diagnose_rewrite_legacy_notes.py -v` — 27/27 pass (14 existing + 13 new)
- [x] `make check` — lint + typecheck + full test suite, all green
- [x] Live smoke test against staging — flag works end-to-end, found one residual doc the original cleanup missed

## Cross-references

- #176 / #180 — original `rewrite-legacy-notes` work where the selector gap was discovered
- #178 — renderer body-only + strict-mode (open follow-up, unrelated to selector)
- #179 — write-call-shape regression test (merged via #183)

### DIFF
--- a/docs/operations/staging-runbook.md
+++ b/docs/operations/staging-runbook.md
@@ -685,6 +685,46 @@ grep -v '"source_url": "http://<bad-host>/' \
   > ${INFLUX_STATE_PATH}/unresolved-slug-collisions.jsonl
 ```
 
+## 8f. Rewriting legacy-shape notes (issues #176 / #181)
+
+A historical writer bug (resolved by 2026-05-08, residue identified
+during the #176 investigation) left some Influx-authored notes in
+Lithos with empty doc-level `tags` / `source_url` / `confidence`
+despite their `content` carrying the correct values inside an
+embedded YAML frontmatter block. The
+`./scripts/influx-diagnose.py rewrite-legacy-notes` subcommand
+recovers these docs by parsing the embedded frontmatter and
+re-issuing the write through `lithos_write` with the structured
+fields as proper API parameters.
+
+### Selector modes
+
+| Selector | When to use |
+| -------- | ----------- |
+| Default (no flags) | Reads `${INFLUX_STATE_PATH}/unresolved-slug-collisions.jsonl` and dedupes the doc ids. Catches the docs actively blocking the slug-collision recovery chain — the operational triage shape. |
+| `--id <doc-id>` (repeatable) | Targets known doc ids directly. Useful for re-running against a specific pair after an incident, or for the post-mortem cleanup loop. With `--apply`, no extra `--yes` is needed for these ids. |
+| `--corpus-scan` (#181) | Walks the on-disk articles subtree (`$LITHOS_KNOWLEDGE_PATH/articles` or the staging-convention fallback at `$INFLUX_ARCHIVE_PATH/../lithos/knowledge/articles`) and surfaces every Influx-authored doc whose doc-level `source_url` is empty — including recoverable docs that haven't yet collided. The broader cleanup shape. Mutually exclusive with `--id`. |
+
+### Outcome buckets (printed in the per-doc summary line)
+
+| Bucket | Meaning |
+| ------ | ------- |
+| `would_rewrite` | Dry-run: parser found a recoverable doc; `--apply` would write it. |
+| `rewrote` | Apply: doc rewritten successfully (`lithos_write status=updated`). |
+| `partner_owns_url` | Apply: the colliding partner doc in a slug-collision pair already owns this `source_url`, so Lithos correctly rejected the duplicate URL. The current doc remains a zombie but is harmless — the partner is canonical and `_classify_squatter` will resolve future collisions against it. |
+| `skipped_already_fixed` | Doc-level `source_url` is already populated. |
+| `skipped_unparseable` | Content has no `---`-fenced inner frontmatter (typical for #162 source-invalid docs). Route those through `invalid-source` (see §8b) or `squatters --apply --no-require-influx-authored` (see §8). |
+| `refused_non_influx_authored` | Safety guard: `metadata.author != 'influx'`. Never auto-rewrite a non-Influx doc. |
+| `failed` | Read or write raised. Investigate the per-doc reason. |
+
+### Safety contract
+
+Same shape as `cmd_squatters`: dry-run by default; `--apply` requires
+one of `--yes <doc-id>`, `--yes-to-all`, or `--id <doc-id>` (the
+`--id` flag is its own confirmation since it names the targets
+explicitly). Idempotent — re-running after `--apply` is a no-op for
+already-fixed docs.
+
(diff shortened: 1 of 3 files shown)


davesnowdon:
Addressed the review's medium-severity finding on the closing-fence parser.

**The bug:** `_select_legacy_doc_ids_from_corpus` used `text.find("---", 4)` to locate the closing fence. That's a substring match, not a line-anchored one — so a YAML value like `title: 'OmniRobot --- a multi-camera platform'` truncates the frontmatter at the inner `---`, `yaml.safe_load` chokes on the truncated payload, and the broken doc never enters the candidate set. Silent false negative.

**The fix:** delegate to `influx.notes._split_frontmatter` (the canonical line-anchored parser already used by `strip_legacy_frontmatter` further down the rewrite path). Now the selector and the per-doc rewrite use the same parser, with identical edge-case behaviour.

**Regression test:** `test_fence_in_value_does_not_truncate_frontmatter` writes a file with `---` inside a quoted title value (deliberately bypassing `yaml.safe_dump`'s auto-quoting that would have masked the bug shape) and asserts the doc IS selected. Verified by temporarily reverting just the parser change locally — the new test failed with `[] != ['fence-in-value-doc']`; restored the fix, it passes.

Test count: 27 → 28. `make check` green. CI on the new commit will follow.

Issue #177, opened 2026-05-24
Surface invalid_url_rejections and prune backlog: historical Sourcegraph `localhost:5174` leak (62 entries)


davesnowdon:
## Symptom

`unresolved-slug-collisions.jsonl` on staging contains **62 entries** whose `source_url` is `http://localhost:5174/<path>`. They will never be re-fetched or auto-recovered because the upstream URL is now different — they're operationally dead data sitting in the backlog. The associated Lithos doc pairs (~124 docs, paired by Influx's slug-collision suffix-retry shape) carry the same `localhost:5174` URL in their YAML frontmatter `source_url:` field.

Surfaced as a side-finding during the diagnosis of #176.

## Scope

| Dimension | Value |
|---|---|
| Affected profile | `ai-coding` (only) |
| Affected source | `rss` (only) |
| Unique URL paths | 9 (all `/blog/<sourcegraph-post-slug>`) |
| Backlog entries | 62 (collisions × suffix-retry collisions) |
| Time window of ingestion | 2026-05-09 05:39:51Z → 2026-05-10 14:56:39Z (~30h) |

Sample paths (representative):

```
/blog/a-new-era-for-sourcegraph-the-intelligence-layer-for-ai-coding-agents-and-developers
/blog/agentic-coding-is-creating-more-code-more-code-is-creating-a-bigger-need-for-code-search
/blog/cross-repository-code-navigation
/blog/deep-search-goes-ga-now-with-role-based-permissions
/blog/introducing-deep-search
/blog/semantic-code-search-what-it-is-and-how-it-works
/blog/why-code-search-at-scale-is-essential-when-you-grow-beyond-one-repository
```

## Root cause (upstream, since-resolved)

The Sourcegraph blog RSS feed shipped articles whose `<link>` elements pointed at `http://localhost:5174/...` — the dev-preview host of their static-site generator. Influx's RSS adapter trusted the absolute URL as-given and wrote the resulting notes to Lithos. Upstream has since fixed the feed.

## Timeline (from run ledger)

| Date | `invalid_url_rejections_total` | Interpretation |
|---|---|---|
| 2026-05-09 → 2026-05-10 | absent / 0 | Pre-validation; bad URLs leaked through to Lithos writes. |
| 2026-05-12 → 2026-05-17 | **30 / run** (every scheduled `ai-coding` run) | Validation now present; rejecting Sourcegraph feed's bad URLs silently. |
[...]

PR #185, opened 2026-05-24
feat: render body-only content + strict-mode write_note assertion (#178)

### LINKED ISSUE #178: Renderer: emit body-only content and reject embedded frontmatter at the lithos_write boundary

davesnowdon:
## Summary

Closes #178.

Drops the embedded `---`-fenced YAML frontmatter block that the Influx renderer historically emitted at the top of every note's `content` argument, and adds a strict-mode assertion in `LithosClient.write_note` so a future regression that re-introduces the embedded-frontmatter shape fails loudly at the MCP boundary.

This is the last #176-spinoff cleanup item. The renderer change brings Influx into compliance with Lithos spec §5.1 (`content` is "Markdown content without frontmatter"; tags / source_url / confidence / note_type / namespace are first-class API parameters), saves ~200 bytes/note of duplicate state, and stops contaminating lithos-enrich's entity extraction.

## What changed

| Layer | Change |
|---|---|
| **Renderer** (`src/influx/renderer.py`) | Delete `_render_frontmatter` and `_format_confidence`. `render_note` output now starts with `# {title}` and skips the fence prefix. Drop the unused `normalise_url` import. `source_url` / `confidence` kept as params for backward compatibility (dead — a follow-up could remove them once callers are updated). |
| **write_note strict-mode** (`src/influx/lithos_client.py`) | Raise `LithosError` if `content` begins with `---\n` or `---\r\n`. Production paths post-#178 can't trigger it; the repair sweep bypasses `write_note` via direct `call_tool`, and the `rewrite-legacy-notes` subcommand strips frontmatter explicitly. |
| **`parse_note`** (`src/influx/notes.py`) | Accept body-only input AND the legacy frontmatter-prefixed shape. When body-only, `frontmatter_raw` is empty in the returned `ParsedNote`. Both shapes preserve byte-exact `## User Notes` regions. Needed because the post-#178 `doc.content` returned by `lithos_read` no longer has the inner frontmatter. |
| **Tests** | 4 obsolete test classes deleted (testing the removed contract). New `TestBodyOnlyOutput` class. New `TestWriteNoteStrictMode` class with 3 cases. `test_no_frontmatter_fence` flipped to `test_body_only_input_parses_with_empty_frontmatter`. Golden fixture regenerated by stripping the outer frontmatter block. |

## Design decisions

**Option A2 (delete `_render_frontmatter`) over A1 (keep as inspection helper):** the issue body offered both. A1 was the lower-friction landing, A2 the cleaner end-state. A2 wins because the rewrite-legacy-notes subcommand already has its own frontmatter-rendering needs covered by ad-hoc YAML construction (it parses *existing* inner frontmatter for migration; doesn't render new ones).

**Strict-mode at `write_note`, not `call_tool`:** The repair sweep and rewrite job both bypass `write_note` and call `call_tool("lithos_write", args)` directly. Putting the assertion in `write_note` covers the live-ingest path (run.py:602) without disrupting the bypass paths that are already legitimately compliant.

**Keep deprecated `source_url` / `confidence` params on `render`:** Removing them is a breaking API change for every source-builder call site. Out of scope here. Filed as a separate cleanup. They're harmless until then — just unused kwargs.

## Test plan

- [x] `uv run pytest tests/ -q` — all green
- [x] `make check` — lint + typecheck + tests, all green locally
- [x] Manually verified render output starts with `# {title}` and contains no `---` fence in the leading region
- [x] Manually verified `write_note` rejects content starting with `---\n` and `---\r\n` via the new tests

## Smoke test deferred

A live staging smoke test (trigger a manual run, observe the on-disk file shape for new docs) would confirm end-to-end behaviour but isn't blocking — the unit tests cover the contract, and we've been operating on staging all night without issue.

## Cross-references

- #176 / #180 — the bug residue investigation that surfaced the spec violation
- #177 / #182 — sibling URL-validation work
- #179 / #183 — write-call-shape regression test
- #181 / #184 — corpus-scan selector

### DIFF
--- a/src/influx/lithos_client.py
+++ b/src/influx/lithos_client.py
@@ -787,7 +787,27 @@ class LithosClient:
         user-notes preservation + retry once, AC-05-E).
         Returns a :class:`WriteResult` so callers can inspect the
         outcome and increment counters (e.g. ``dedup_skipped``).
+
+        Strict-mode contract (#178): ``content`` must be body-only
+        markdown — no leading ``---``-fenced YAML frontmatter block.
+        Lithos owns the outer frontmatter and persists ``tags`` /
+        ``source_url`` / ``confidence`` / ``note_type`` / ``namespace``
+        from the API parameters on this call (spec §5.1).  A
+        ``LithosError`` is raised if ``content`` starts with ``---\\n``
+        so a renderer regression that re-introduces the embedded
+        frontmatter shape fails loudly at the boundary instead of
+        silently doubling the on-disk state.
         """
+        if content.startswith("---\n") or content.startswith("---\r\n"):
+            raise LithosError(
+                "content begins with a '---' frontmatter fence — "
+                "the lithos_write contract requires body-only content; "
+                "tags / source_url / confidence / note_type / namespace "
+                "must be passed as API parameters, not inlined in content "
+                "(spec §5.1, issue #178)",
+                operation="write_note",
+                detail="embedded_frontmatter",
+            )
         args: dict[str, Any] = {
             "title": title,
             "content": content,
--- a/src/influx/lithos_client.py
+++ b/src/influx/lithos_client.py
@@ -787,50 +787,13 @@ class LithosClient:
         user-notes preservation + retry once, AC-05-E).
         Returns a :class:`WriteResult` so callers can inspect the
         outcome and increment counters (e.g. ``dedup_skipped``).
-
-        Strict-mode contract (#178): ``content`` must be body-only
-        markdown — no leading ``---``-fenced YAML frontmatter block.
-        Lithos owns the outer frontmatter and persists ``tags`` /
-        ``source_url`` / ``confidence`` / ``note_type`` / ``namespace``
-        from the API parameters on this call (spec §5.1).  A
-        ``LithosError`` is raised if ``content`` starts with ``---\\n``
-        so a renderer regression that re-introduces the embedded
-        frontmatter shape fails loudly at the boundary instead of
-        silently doubling the on-disk state.
-
-        Canonical-URL normalization (FR-MCP-4): ``source_url`` is
-        normalised via :func:`influx.urls.normalise_url` before being
-        forwarded to ``lithos_write``.  Pre-#178 this happened inside
-        the rendered frontmatter; after the renderer change, the
-        ``write_note`` boundary owns it so callers don't have to
-        pre-normalise.  Lithos's internal ``normalize_url`` would
-        canonicalise for dedup-map storage anyway, but normalising
-        here keeps Influx's own logs / telemetry / ``WriteResult``
-        in canonical form and stops dedupe identity from drifting
-        for any caller that passes a tracking-param-laden URL.
         """
-        if content.startswith("---\n") or content.startswith("---\r\n"):
-            raise LithosError(
-                "content begins with a '---' frontmatter fence — "
-                "the lithos_write contract requires body-only content; "
-                "tags / source_url / confidence / note_type / namespace "
-                "must be passed as API parameters, not inlined in content "
-                "(spec §5.1, issue #178)",
-                operation="write_note",
-                detail="embedded_frontmatter",
-            )
-        # FR-MCP-4: canonicalise source_url at the API boundary.  See
-        # ``_safe_normalise_url`` — never crash the write loop on a
-        # malformed URL, fall through with the raw value so the write
-        # at least attempts (Lithos will reject it as invalid_input
-        # if it really is unusable).
-        canonical_source_url = _safe_normalise_url(source_url)
         args: dict[str, Any] = {
             "title": title,
             "content": content,
             "agent": agent,
             "path": path,
-            "source_url": canonical_source_url,
+            "source_url": source_url,
             "tags": list(tags),
             "confidence": confidence,
             "note_type": note_type,
@@ -839,25 +802,25 @@ class LithosClient:
         if expires_at is not None:
             args["expires_at"] = expires_at
         result = await self.call_tool("lithos_write", args)
-        parsed = self._parse_write_response(result, source_url=canonical_source_url)
+        parsed = self._parse_write_response(result, source_url=source_url)
 
         if parsed.status == "slug_collision":
             return await self._retry_slug_collision(
-                args, source_url=canonical_source_url, initial_collision=parsed
+                args, source_url=source_url, initial_collision=parsed
             )
 
         if parsed.status == "version_conflict":
             return await self._retry_version_conflict(
                 args,
                 note_id=parsed.detail,
-                source_url=canonical_source_url,
+                source_url=source_url,
                 original_tags=tags,
             )
 
         if parsed.status == "content_too_large":
             return await self._retry_content_too_large(
                 args,
-                source_url=canonical_source_url,
+                source_url=source_url,
                 original_tags=tags,
             )
 
--- a/src/influx/lithos_client.py
+++ b/src/influx/lithos_client.py
@@ -797,6 +797,17 @@ class LithosClient:
         so a renderer regression that re-introduces the embedded
         frontmatter shape fails loudly at the boundary instead of
         silently doubling the on-disk state.
+
+        Canonical-URL normalization (FR-MCP-4): ``source_url`` is
+        normalised via :func:`influx.urls.normalise_url` before being
+        forwarded to ``lithos_write``.  Pre-#178 this happened inside
+        the rendered frontmatter; after the renderer change, the
+        ``write_note`` boundary owns it so callers don't have to
+        pre-normalise.  Lithos's internal ``normalize_url`` would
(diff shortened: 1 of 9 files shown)


davesnowdon:
Addressed the review's medium-severity finding on source_url canonicalisation.

**The bug:** Pre-#178, the renderer normalised `source_url` via `influx.urls.normalise_url` inside the rendered frontmatter (FR-MCP-4). After the renderer change, the frontmatter is gone and `write_note` was forwarding the raw URL straight to `lithos_write`. Lithos's internal `normalize_url` still covers dedup-map correctness on receipt, so no user-visible breakage today — but Influx's own logs, telemetry, `WriteResult.source_url`, slug-collision detail strings, and any caller consuming the raw URL silently shifted to the un-canonical shape. Identity drift between Influx and Lithos.

**The fix:** canonicalise at the `write_note` boundary using the existing `_safe_normalise_url` helper (defensive — falls back to raw on malformed URL, never crashes the write loop). The canonical URL flows through to:
- the MCP `args["source_url"]`
- `_parse_write_response`'s `source_url=` (so `WriteResult` is canonical)
- all three retry paths (`_retry_slug_collision`, `_retry_version_conflict`, `_retry_content_too_large`)

**Regression test:** `test_canonicalises_source_url_at_the_api_boundary` — pass in the reviewer's exact example URL (`https://ArXiv.org:443/abs/2601.12345?utm_source=twitter&utm_id=42`), assert both the MCP `call_tool` args AND the returned `WriteResult.source_url` are the canonical form. Verified by reverting just the args-dict line — test failed with the raw URL; restored, test passes.

Test count: 27 → 28 in `test_lithos_client.py`. `make check` green. CI on the new commit will follow.

davesnowdon:
Heads-up on the CI run that just failed — the 8 failing tests are all in `tests/integration/test_arxiv_pipeline_end_to_end.py` and assert `lithos_write` is called, but the pipeline silently completes with 0 writes captured by the fake server. The captured log shows `PermissionError: '/state'` from `run_ledger` trying to mkdir the default state dir.

**Pre-existing on main, not introduced by this PR.** Verified locally:

```
# On main (no edits):
$ env -i HOME="$HOME" PATH="$PATH" bash -c 'uv run --quiet pytest tests/integration/test_arxiv_pipeline_end_to_end.py::TestRunProfileDrivesExtraction::test_lithos_write_carries_text_html_and_full_text'
FAILED ... assert 0 == 1
1 failed, 2 warnings in 0.92s

# On feat/178-renderer-body-only:
FAILED ... assert 0 == 1
1 failed, 2 warnings in 0.92s
```

Same failure shape on both. My PR doesn't touch the arXiv pipeline or run_ledger.

The same test passed in CI on #184 (the most recent main merge, 22:24 UTC yesterday) — the runner environment seems to have shifted between then and now. Either the CI image was updated, `/state` writability differs, or there's a non-deterministic interaction.

I've pushed an empty commit to retrigger CI. If the retry still fails on these tests, I'll file a separate issue tracking the `/state`-writability assumption (the test fixture's default `state_dir = '/state'` from `tests/conftest.py:24` only works when the runner can mkdir `/state`).

The PR's actual changes — renderer body-only, write_note strict-mode, parse_note body-tolerance, source_url canonicalisation at the API boundary — all have green unit-test coverage and are unrelated to the arXiv-pipeline failure mode.

davesnowdon:
**Root cause found — date-roll fixture rot, not this PR.**

`tests/integration/test_arxiv_pipeline_end_to_end.py:60` hardcodes the simulated ArXiv entry's published date:

```xml
<published>2026-04-25T00:00:00Z</published>
```

The test config uses `lookback_days=30`. Today is **2026-05-25**, exactly 30 days after the fixture date. The lookback filter rejects, the pipeline filters out the only candidate, 0 writes happen, the assertion `len(write_calls) == 1` blows up.

Yesterday (2026-05-24, when #184's main CI run passed at 22:24 UTC), the gap was 29 days — inside the window. Today, the gap is 30 — at the boundary, filtered out.

Bisect evidence (full revert of all my PR's src/ changes to main, test still fails locally):

```
$ git checkout main -- src/influx/renderer.py src/influx/lithos_client.py
$ uv run --quiet pytest tests/integration/test_arxiv_pipeline_end_to_end.py::TestRunProfileDrivesExtraction::test_lithos_write_carries_text_html_and_full_text
FAILED ... assert 0 == 1
1 failed, 2 warnings in 0.90s
```

This affects all 8 failing tests because they share the `_ATOM_FEED` fixture's published date.

**Suggested fix (separate PR, out of scope for #178):** either bump the fixture date dynamically (e.g. `datetime.now() - timedelta(days=1)`) or wrap each affected test in a freezer/frozen-time fixture that pins the wall clock to a known date inside the lookback window.

**For this PR:** the 8 failing tests are unrelated to the renderer/strict-mode/source_url-canonicalisation changes. CI green on every other test (2338/2346 passing). Happy to either:

1. **Merge with the date-roll failures noted** — they'll bite every PR equally until the fixture is bumped.
2. **Stack a tiny fixture-bump commit** here so this PR's CI is fully green, then file the proper fix as a follow-up.
3. **Wait for a separate PR** that lands the fixture fix on main first, then rebase this PR.

Operator's call.

PR #251, opened 2026-07-06
refactor(notes): give CanonicalNote a single owning module (PR 1/6)

davesnowdon:
## Context

From the 2026-07-06 architecture review (finding 1 / Lithos task `9ae1086e`). The **CanonicalNote** is Influx's central artifact, but no module owned its shape: the renderer emitted section order imperatively (no `SECTION_ORDER` constant existed), `notes.py` parsed it, `lithos_client` and the repair sweep each hand-spliced sections with their own heading matchers, and the byte-exact `## User Notes` invariant was implemented ~5× with 3 different matching semantics (a CRLF-tolerant regex, an unanchored `str.find` that could truncate content mid-line, and a `\b` regex).

This is **PR 1 of 6** — the foundation. It introduces the owning module and delegates the existing parser to it, but **no rewrite call site migrates yet** (that's PRs 2–5).

## What this PR does

- **New `src/influx/canonical_note.py`** — the single owner of the note shape. Imports only Foundation (`errors`, `schemas`). Owns:
  - `SECTION_ORDER` as data (+ `TIER2_SECTIONS`/`TIER3_SECTIONS` and section-name constants)
  - the **one** line-anchored `## User Notes` matcher (`find_user_notes_start`, `split_user_notes`)
  - `Section` / `CanonicalNote` with immutable `get_section`/`upsert_section`/`drop_sections`, and `parse`/`parse_lenient`/`serialize`
  - the string-level rewrite ops PRs 2–5 will swap onto: `insert_full_text_section`, `render_tier3_sections`, `insert_tier3_sections`, `drop_tier2`, `drop_tier2_and_tier3`, `extract_section_body`, `upsert_archive_path`, `replace_profile_relevance_section`, `upsert_section_text`, `graft_user_notes`
  - `ProfileRelevanceEntry` + de-privatized `render_profile_relevance_body`
- **`notes.py` delegates** — `ParsedNote = CanonicalNote`, `ParsedSection = Section`, `parse_note = parse`; parse internals moved verbatim; the wrong-direction `notes → renderer` import is removed.
- **`renderer.py`** re-exports `ProfileRelevanceEntry` and keeps a `_render_profile_relevance_body` alias so the not-yet-migrated `lithos_client` import still resolves.
- **`scripts/influx-diagnose.py`** repointed (6 sites) to the moved `_split_frontmatter`/`_split_title` in `canonical_note`.
- **Registered** `influx.canonical_note` in both import-linter contracts and `docs/architecture.toml` (LithosBridge). The generated component diagram does not drift.

## Migration behaviour: parity helpers vs. intentional fixes

**Most** ops are byte-conservative — verified byte-identical to today's `lithos_client` / `repair_hooks` / `repair_counters` helpers on canonical inputs, pinned by the transitional `TestLegacyParityTransitional` suite (deleted as each legacy helper is removed). The only across-the-board tightening is the **locator**: every op matches a heading as a whole line, which diverges from the legacy helpers *only* on pathological input (a mid-line `## User Notes` literal, a `### User Notes` H3) — and there the change is protective (it stops the old `str.find` from truncating a note mid-content).

**Three ops intentionally fix known legacy edge behaviour** when their call site migrates (so they are *excluded* from the parity suite, and PRs 3/4 update the corresponding contract-test assertions deliberately):
- `drop_tier2` / `drop_tier2_and_tier3` — preserve a trailing `## User Notes` region **byte-exactly** instead of applying the legacy whole-document `rstrip()` that trimmed user-note trailing whitespace (lands in PR 3, oversize-trim path).
- `upsert_archive_path` — treats a `path:` line **anywhere** in the `## Archive` body (any indentation) as already present, so a hand-edited section never gets a duplicate `path:` (lands in PR 4, archive repair).

## Test plan

- `tests/unit/test_canonical_note.py` (63 tests) + `tests/fixtures/canonical_notes/` corpus: round-trip identity `serialize(parse(x)) == x`, per-op `## User Notes` byte-exactness (incl. trailing whitespace), the matcher divergence matrix, CRLF `extract_section_body`, `upsert_archive_path` idempotence across metadata/indented/blank-line/other-section cases, documented CRLF-legacy serialize normalisation, and the transitional legacy-parity suite.
- Existing regression net untouched and green: `test_notes_parse.py`, `test_user_notes_roundtrip.py`, `test_multi_profile_merge.py`, `test_repair_counters.py`, the `test_lithos_client.py` contract suite, `tests/meta/*` (import-linter contract, orphan check, diagram drift).
- `make check`: lint + pyright clean; the whole suite passes on CI. Locally one test (`test_cli_http.py::test_run_conflict_exit_1`, a `python -m influx run` subprocess that times out at 10s) fails but reproduces identically on `main` with this branch stashed — pre-existing environmental flake, confirmed passing on this PR's CI.

## Follow-ups (this stack)

PR 2 migrate `repair_hooks` onto the shared `render_tier3_sections`/insert helpers · PR 3 migrate `lithos_client` merge/trim (the User Notes matcher + drop-op bytes change) · PR 4 `repair.py` archive splice (`upsert_archive_path`) + `parse_lenient` · PR 5 `repair_counters` placement · PR 6 (optional) renderer composes via `serialize` + drop dead `source_url` param.


davesnowdon:
Thanks — this was a sharp review; three of the four are real foundation bugs and I've fixed them in `2198b5e` rather than carry them into PRs 2–5. Answering the questions directly:

**Q1 — byte-exact `## User Notes` vs legacy `.rstrip()`?** Byte-exact. Since this module *owns* the invariant, the drop ops now preserve the trailing `## User Notes` region exactly and no longer apply the legacy whole-document `rstrip()`. Like the `## User Notes` matcher tightening, this is an intended divergence from the current `lithos_client._drop_tier2/_drop_tier2_and_tier3` — the changed bytes reach production only when **PR 3** migrates the oversize-trim path onto these helpers (I've noted it in the helper docstring and the parity suite so PR 3 updates the contract-test assertion deliberately). New test asserts byte-exact preservation including trailing spaces/blank lines in the region.

**Q2 — `upsert_archive_path` robustness?** Made robust. It now scans the whole `## Archive` section body (bounded to the next `## ` heading) for an existing line-anchored `path:` and no-ops if one exists anywhere — so a hand-edited section with other metadata before `path:` never gets a duplicate (which `parse_archive_path` would then reject). Tests cover metadata-before-path idempotence, insert-once-when-metadata-but-no-path, and a `path:` in a *later* section not suppressing insertion.

**Q3 — `serialize` and CRLF/legacy?** Canonical LF renderer output only. I've made the docstring explicit: `serialize(parse(x)) == x` is promised for LF renderer output; CRLF/legacy notes round-trip with separators normalised to LF (the `## User Notes` region alone stays byte-exact via `parse`). Added a test pinning that normalisation as intentional.

**Finding 2 — `extract_section_body` CRLF:** fixed to skip a `\r\n` (not just `\n`) after the heading; CRLF test added.

On the parity point: rather than a broad legacy-parity harness (inappropriate now that drop/archive ops intentionally diverge), I committed a small **transitional** `TestLegacyParityTransitional` covering only the ops that must stay byte-identical for their migration — `render_tier3_sections`, `insert_full_text_section`, `insert_tier3_sections`, `insertion_point`, `graft_user_notes`, `replace_profile_relevance_section`, `upsert_section_text` vs their current `repair_hooks`/`lithos_client`/`repair_counters` counterparts. These get deleted as each legacy helper is removed in its migration PR. The drop/archive contracts are pinned directly (byte-exact / idempotence) instead of by equivalence to the buggy legacy.

`make check` (minus the pre-existing `test_run_conflict_exit_1` subprocess flake, which passed on this PR's CI) is green; 62 canonical_note tests. Ready for another look.

davesnowdon:
Thanks for the re-review — both minor items are addressed in `646aa7b`.

**Q1 — indented `path:`?** Yes, treated as present. You're right that `parse_archive_path` strips the section body before matching, so an indented lone `  path:` line is a valid archive path — and the old `^path:` would have inserted a duplicate. Confirmed it (`path:` count went to 2), then changed `_PATH_LINE_RE` to `^[ \t]*path:` and added `test_idempotent_when_path_is_indented`. Detection now matches the parser's tolerance.

**Q2 — distinguish parity helpers from intentional fixes?** Done, in both places. The module docstring now separates the byte-conservative helpers (preserve legacy bytes; only the heading *locator* tightens, and only on pathological input) from the three that intentionally fix legacy edge behaviour on migration — `drop_tier2`, `drop_tier2_and_tier3` (byte-exact User Notes) and `upsert_archive_path` (whole-section `path:` idempotence). The PR description's "Migration behaviour" section now makes the same split explicit and notes those three are excluded from the transitional parity suite, with PRs 3/4 updating the contract-test assertions deliberately.

63 canonical_note tests; ruff + pyright clean. Thanks again for catching the indentation edge — that one would have produced malformed Archive sections after the PR 4 migration.

PR #252, opened 2026-07-06
refactor(repair): migrate repair_hooks + renderer onto shared canonical ops (PR 2/6)

davesnowdon:
## Context

**PR 2 of 6** in the CanonicalNote stack (finding 1 / Lithos task `9ae1086e`), on top of #251. PR 1 introduced `influx.canonical_note` as the single owner of the note shape but migrated no call sites. This PR migrates the first two: the **repair sweep** and the **renderer**, eliminating the byte-for-byte Tier 3 duplication the review flagged.

## What this PR does

- **`repair_hooks.py`** — deletes the local section-shape helpers (`_find_insertion_point`, `_insert_full_text_section`, `_render_tier3_sections`, `_insert_tier3_sections`, `_extract_full_text_body`) and their now-unused regexes, routing the Tier 2 / Tier 3 hooks through `canonical_note.insert_full_text_section` / `insert_tier3_sections` / `extract_section_body`. `_extract_title` stays local (no canonical equivalent). Net −~70 lines of duplicated splice logic.
- **`renderer.py`** — the Tier 3 block (`## Claims` / `## Datasets & Benchmarks` / `## Builds On` / `## Open Questions`) now calls `canonical_note.render_tier3_sections`, the single definition it previously duplicated **byte-for-byte** with `repair_hooks._render_tier3_sections`.
- **Diagram** — `repair_hooks` was the only Repair-tier module importing `influx.schemas` (`Tier3Extraction`); dropping that import removes the `Repair → Schemas` edge. Regenerated diagram committed so the `diagrams` gate stays green.

## Byte-identity

The renderer output is byte-identical (verified incl. the empty-Tier-3-list edge via golden-fixture round-trip; `test_tier3_render` and `test_user_notes_roundtrip` pin the exact bytes). The repair sweep's `SweepHooks` test-injection seam is untouched — the repair-sweep integration tests drive the same public `sweep()` and pass unchanged.

## Tests

- The Tier 3 / full-text section-op coverage that lived in `test_repair_hooks_production` (extract/insert/render) moves to `test_canonical_note`, where the ops now live; added empty-list and insert-ordering cases.
- Removed the four `repair_hooks` entries from the transitional legacy-parity suite now that those helpers are gone (the `lithos_client` / `repair_counters` parity cases remain for PRs 3/5).

## Follow-ups (this stack)

PR 3 migrate `lithos_client` merge/trim (the User Notes matcher + drop-op bytes change) · PR 4 `repair.py` archive splice (`upsert_archive_path`) + `parse_lenient` · PR 5 `repair_counters` placement · PR 6 (optional) renderer composes via `serialize` + drop dead `source_url` param.

### DIFF
--- a/docs/generated/architecture.md
+++ b/docs/generated/architecture.md
@@ -77,7 +77,6 @@ graph TD
   Repair --> Enrich
   Repair --> LithosBridge
   Repair --> Observability
-  Repair --> Schemas
   Repair --> Sources
   Repair --> Storage
   Sources --> Common
--- a/src/influx/renderer.py
+++ b/src/influx/renderer.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 from influx.canonical_note import (
     ProfileRelevanceEntry,
     render_profile_relevance_body,
+    render_tier3_sections,
 )
 from influx.errors import InfluxError
 from influx.schemas import Tier1Enrichment, Tier3Extraction
@@ -218,20 +219,11 @@ def render_note(
     if full_text:
         output += f"\n## Full Text\n{full_text}\n"
 
-    # Tier 3 sections (US-012) — omitted entirely when absent (FR-ENR-6)
+    # Tier 3 sections (US-012) — omitted entirely when absent (FR-ENR-6).
+    # The section block is owned by canonical_note (shared with the repair
+    # sweep); the leading blank-line separator is added here.
     if tier3_extraction is not None:
-        output += "\n## Claims\n"
-        for claim in tier3_extraction.claims:
-            output += f"- {claim}\n"
-        output += "\n## Datasets & Benchmarks\n"
-        for ds in tier3_extraction.datasets:
-            output += f"- {ds}\n"
-        output += "\n## Builds On\n"
-        for item in tier3_extraction.builds_on:
-            output += f"- {item}\n"
-        output += "\n## Open Questions\n"
-        for q in tier3_extraction.open_questions:
-            output += f"- {q}\n"
+        output += "\n" + render_tier3_sections(tier3_extraction)
 
     # Profile Relevance section — always emitted to keep the canonical
     # note shape stable (US-007); body is empty when no entries are given.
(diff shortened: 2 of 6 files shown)


davesnowdon:
Thanks — added the exact-byte tests in `943f619`.

**Recommended item (exact-output insertion tests):** replaced the order/substring assertions with a `TestSectionInsertBytes` class that pins, for `insert_full_text_section` and `insert_tier3_sections`:
- the precise inserted bytes and blank-line shape (byte-identical to the legacy `repair_hooks` helpers this PR replaced — including the `\n\n\n` before an inserted section that follows an existing one; preserved deliberately, not cleaned up, to keep the migration byte-exact);
- the full insertion-point fallback chain — before `## Profile Relevance`, fallback before `## User Notes` when PR is absent, fallback at EOF when neither exists;
- byte-exact preservation of the surrounding note, asserting the `## User Notes` region (incl. trailing whitespace) is unchanged.

**Your question — is the stricter canonical heading matching intended production behaviour for the repair hooks?** Yes, intended, and it's a no-op on real sweep input. The old `repair_hooks` insertion matchers were already line-anchored (`^## Profile Relevance\b`, `^## User Notes\b`, `^## Full Text[ \t]*$`); the canonical locators are whole-line (`^## <heading>[ \t]*(?=\r?\n|$)`). On canonical notes (what the sweep actually processes) they resolve to the identical positions — the parity checks in PR 1 confirmed byte-identity. They diverge only on non-canonical/hand-edited input: canonical won't treat `## Profile Relevance: foo` as the heading (protective, matches what the parser already does), and it *gains* CRLF tolerance for `## Full Text` extraction that the old `$`-anchored regex lacked. So it's the same documented "locator tightening" from PR 1, now reaching the repair hooks — behaviour-preserving on production notes, protective on pathological ones. I've left the PR body's "byte-identical migration" framing since it's accurate for the inputs the sweep sees; happy to add a sentence calling out the pathological-input divergence explicitly if you'd prefer.

`make check` clean (minus the known `test_run_conflict_exit_1` subprocess flake); 65 canonical_note tests.

PR #253, opened 2026-07-06
refactor(lithos): migrate merge/trim onto shared canonical ops (PR 3/6)

davesnowdon:
## Context

**PR 3 of 6** in the CanonicalNote stack (finding 1 / Lithos task `9ae1086e`), on top of #251 and #252. This is the step the plan flagged as **riskiest and deliberately isolated**: it deletes `lithos_client`'s hand-rolled section splicers and routes the write-retry paths through `influx.canonical_note` — which is where the section-locator tightening and the byte-exact `## User Notes` invariant finally reach production.

## What this PR does

- **`lithos_client.py`** — deletes `_preserve_user_notes`, `_replace_profile_relevance_section`, `_drop_tier2`, `_drop_tier2_and_tier3` and their heading-literal constants (~140 lines). The version-conflict merge now grafts User Notes via `canonical_note.graft_user_notes`; the oversize-trim retries call `canonical_note.drop_tier2` / `drop_tier2_and_tier3`; the Profile Relevance merge routes through `canonical_note.replace_profile_relevance_section`. Drops the private `renderer._render_profile_relevance_body` import.
- **`repair.py`** — the two late "avoid circular dependency" imports of `lithos_client` internals (the sweep's version-conflict and chronic-oversize paths) collapse into a single top-level `canonical_note` import. `canonical_note` imports only Foundation, so there's no cycle to dodge; `Repair → LithosBridge` already existed via `notes`, so no diagram change.

## ⚠️ Intended behaviour change (this is the one PR in the stack that changes production bytes)

Both changes were introduced and unit-tested in `canonical_note` back in PR 1; PR 3 is where they go live.

1. **Trim byte-exactness.** The legacy `_drop_tier2` / `_drop_tier2_and_tier3` ended with a whole-document `.rstrip()`, so an oversize retry silently stripped trailing whitespace and blank lines from the final `## User Notes` section. The canonical drop ops preserve the tail byte-exactly — the User Notes invariant now holds on the *trim* path, not just on first write.
2. **Locator tightening.** `graft` / `replace` / `drop` use line-anchored heading matchers (`^## Heading[ \t]*(?=\r?\n|$)`) instead of unanchored `str.find`. On canonical notes they resolve to identical positions (verified by PR 1's parity checks). On *pathological* input they are strictly protective: a mid-line `## User Notes` literal, or a `### User Notes` sub-heading in the incoming content, can no longer be mistaken for the section boundary and truncate the note during merge.

Neither case affects a normally-rendered note; both are documented in `canonical_note`'s module docstring.

## Tests

- The contract suite (merge + trim vs `FakeLithosServer`) stays green unchanged — its assertions are substring/structural, so the trailing-whitespace fix doesn't perturb them.
- **Added** a contract test (`test_first_retry_preserves_user_notes_bytes`) pinning byte-exact `## User Notes` preservation through the Tier-2 trim retry — the clearest proof the fix reaches the write path (legacy would have truncated `Keep this.  \n\n` to `Keep this.`).
- Repointed the `_replace_profile_relevance_section` unit + integration tests to `canonical_note.replace_profile_relevance_section`.
- Removed the two now-obsolete `lithos_client` legacy-parity cases from `test_canonical_note` (helpers gone); the `repair_counters` parity case remains for PR 5.

## Follow-ups (this stack)

PR 4 `repair.py` archive splice (`upsert_archive_path`) + `parse_lenient` · PR 5 `repair_counters` placement onto `insertion_point`/`upsert_section_text` · PR 6 (optional) renderer composes via `serialize` + drop dead `source_url` param.

### DIFF
--- a/src/influx/lithos_client.py
+++ b/src/influx/lithos_client.py
@@ -25,6 +25,12 @@ from mcp.client.session import ClientSession
 from mcp.client.sse import sse_client
 from mcp.shared.exceptions import McpError
 
+from influx.canonical_note import (
+    drop_tier2,
+    drop_tier2_and_tier3,
+    graft_user_notes,
+    replace_profile_relevance_section,
+)
 from influx.dedup import compose_dedup_query
 from influx.errors import ConfigError, LCMAError, LithosError
 from influx.notes import (
@@ -36,11 +42,7 @@ from influx.notes import (
     merge_tags as _canonical_merge_tags,
 )
 from influx.renderer import (
-    ProfileRelevanceEntry,
     merge_profile_relevance_union,
-)
-from influx.renderer import (
-    _render_profile_relevance_body as _render_pr_body,
 )
 from influx.urls import normalise_url
 
@@ -435,29 +437,6 @@ def _merge_tags(existing_tags: list[str], new_tags: list[str]) -> list[str]:
     return _canonical_merge_tags(existing_tags=existing_tags, new_tags=new_tags)
 
 
-_USER_NOTES_MARKER = "## User Notes"
-
-
-def _preserve_user_notes(existing_content: str, new_content: str) -> str:
-    """Merge content, preserving ``## User Notes`` from the existing note.
-
-    The ``## User Notes`` section and everything beneath it in
-    *existing_content* replaces any ``## User Notes`` already present
-    in *new_content* (AC-05-E).
-    """
-    idx = existing_content.find(_USER_NOTES_MARKER)
-    if idx == -1:
-        return new_content
-    user_notes_block = existing_content[idx:]
-
-    new_idx = new_content.find(_USER_NOTES_MARKER)
-    base = new_content[:new_idx].rstrip() if new_idx != -1 else new_content.rstrip()
-    return base + "\n\n" + user_notes_block
-
-
-_PROFILE_RELEVANCE_MARKER = "## Profile Relevance"
-
-
 def _merge_profile_relevance_in_content(
     existing_content: str,
     new_content: str,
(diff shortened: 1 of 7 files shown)


davesnowdon:
Thanks — both test-gap findings addressed in `928affe`, and answers to your two questions below.

**Finding 1 (repair sweep byte preservation).** Added `test_trim_retries_preserve_user_notes_bytes` to the chronic-oversize suite. It drives a chronic note through the real `sweep()` so `_rewrite_sweep_note`'s Tier-2 and Tier-1-only trims run via `canonical_note.drop_tier2` / `drop_tier2_and_tier3`, then asserts both retry payloads (`write_calls[1]` and `write_calls[2]`) preserve `## User Notes\nKeep verbatim.  \n\n\n` byte-exactly — trailing spaces and blank lines intact — while Full Text (and, on attempt 3, Claims) are gone. This works cleanly because the default `SweepHooks()` has all hooks `None`, so no stage mutates `note["content"]` (the counted-failure content bump at repair.py:574 is guarded by `classify_failure == "counted"` and never fires here) — the trim runs on the seeded content verbatim, which is what makes the exact-byte assertion deterministic.

**Finding 2 (version-conflict locator tightening).** Added `test_version_conflict_locator_ignores_midline_user_notes`. The incoming body carries a mid-line `## User Notes` literal *before* the real heading (`# Summary\nSee the ## User Notes section below.\n\n## User Notes\n`); the existing note's region has trailing whitespace. It asserts the retry content (a) keeps the mid-line literal (`See the ## User Notes section below.`) — the legacy unanchored `str.find` would have truncated the base right there — and (b) ends with the existing region byte-exactly. That pins the tightened locator's protective behaviour at the production retry boundary.

**Q1 — do I intend the sweep to have the same production-boundary coverage as `write_note`?** Yes, and now it does (Finding 1 above). The sweep and the client share the exact same `drop_tier2` / `drop_tier2_and_tier3` functions, so the canonical unit tests already covered the transform; the gap was purely "is it still wired correctly through the sweep," which the new workflow test guards.

**Q2 / Finding 3 — is `replace_profile_relevance_section` intentionally keeping the legacy `find("\n## ")` end-boundary?** Yes, deliberately, for this PR. That function is a byte-for-byte port of the legacy `lithos_client._replace_profile_relevance_section` (PR 1 pinned the parity), and the Profile Relevance merge is on the multi-profile write path — I don't want to fold a second, byte-sensitive change into the same PR that already carries the stack's one intentional bytes change. You're right that `find("\n## ")` for the end boundary is slightly off-story vs the anchored start matcher; in practice they resolve identically here (the search starts immediately after the `## Profile Relevance` heading text, so there's no position-0 heading to disagree about), but I agree it should converge on the shared section-span helper. I'll take that as a focused item in **PR 6** (the renderer/cleanup PR), with a parity test asserting byte-identity before and after — rather than as an unpinned change here. Flagging it now so it doesn't get lost.

`make check` was green before these test-only additions (minus the known `test_run_conflict_exit_1` subprocess flake); the two new tests pass locally and lint/format clean.

PR #254, opened 2026-07-06
refactor(repair): migrate archive splice + note parse onto canonical ops (PR 4/6)

davesnowdon:
## Context

**PR 4 of 6** in the CanonicalNote stack (finding 1 / Lithos task `9ae1086e`), on top of #251–#253. Migrates `repair.py`'s two remaining pieces of hand-rolled note-structure code onto `influx.canonical_note`.

## What this PR does

- **`_run_archive_retry`** — the raw `## Archive` splice (`find("## Archive\n")` + `startswith("path:")` insert) becomes `canonical_note.upsert_archive_path`. On canonical LF notes this is byte-identical, but it fixes two latent bugs on the archive-download retry path:
  - **CRLF** — the legacy literal `find("## Archive\n")` never matches a `## Archive\r\n` heading, so a CRLF note's freshly downloaded archive path was silently dropped. The anchored matcher is CRLF-tolerant.
  - **Idempotence** — the legacy check only inspected the byte immediately after the heading, so a `path:` line preceded by a blank line or indented would be duplicated. `upsert_archive_path` scans the whole `## Archive` section.
- **`_process_sweep_note`** — `parse_note(_content_for_note_parse(note))` becomes `parse_lenient(str(note.get("content") or ""), fallback_title=_doc_title(note))`. `_content_for_note_parse` re-implemented the parser's title detection to reattach the doc-level title that `read_note` strips; `parse_lenient` owns that now via `_has_title_heading`, which mirrors the real parser's `_split_title` (the removed local copy used `str.splitlines`, subtly divergent on form-feed characters). `_doc_title` stays in `repair.py` — note-dict title resolution is repair-domain, not note-shape.

## Tests

- **Added** `test_archive_download_splices_path_into_crlf_note` — drives an archive-download sweep on a CRLF-shaped note and asserts the path is spliced below the `## Archive\r\n` heading at the production boundary (the exact case the legacy `find` missed; the inserted line is LF-terminated, matching the canonical op).
- **Added** `_has_title_heading` indented-H1 / H2-exclusion cases to `test_canonical_note` — the title-detection nuance that moved out of `repair.py`.
- Replaced `repair.py`'s `TestContentForNoteParse` with `TestSweepNoteParse`, covering the `parse_lenient` + `_doc_title` composition and the `None`-content coercion; the granular byte-shape cases now live with `parse_lenient` in `test_canonical_note`.
- `upsert_archive_path`'s own idempotence/CRLF behaviour was already unit-tested in `test_canonical_note` (PR 1).

## Follow-ups (this stack)

PR 5 `repair_counters` placement onto `insertion_point` / `upsert_section_text` · PR 6 (optional) renderer composes via `serialize`, drop the dead `source_url` param, and converge `replace_profile_relevance_section`'s end-boundary onto the shared section-span helper (carried over from the #253 review).

### DIFF
--- a/src/influx/repair.py
+++ b/src/influx/repair.py
@@ -28,6 +28,8 @@ from influx.canonical_note import (
     drop_tier2,
     drop_tier2_and_tier3,
     graft_user_notes,
+    parse_lenient,
+    upsert_archive_path,
 )
 from influx.errors import ExtractionError, LCMAError, LithosError
 from influx.notes import merge_tags
@@ -1018,20 +1020,13 @@ def _run_archive_retry(
             span_name="influx.repair.archive",
         ):
             downloaded_path = hook(note)
-        # Patch ## Archive with the freshly downloaded path (idempotent —
-        # only inserts when no path: line is already present).
-        content = str(note.get("content", ""))
-        marker = "## Archive\n"
-        idx = content.find(marker)
-        if idx >= 0:
-            insert_pos = idx + len(marker)
-            rest = content[insert_pos:]
-            if not rest.startswith("path:"):
-                note["content"] = (
-                    content[:insert_pos]
-                    + f"path: {downloaded_path}\n"
-                    + content[insert_pos:]
-                )
+        # Patch ## Archive with the freshly downloaded path. Idempotent,
+        # CRLF-tolerant, and scans the whole section for an existing
+        # ``path:`` line — canonical_note owns the splice (a no-op when the
+        # section is absent or already carries a path).
+        note["content"] = upsert_archive_path(
+            str(note.get("content", "")), downloaded_path
+        )
         return current_tags, downloaded_path, True
     except (ExtractionError, LithosError) as exc:
         failure_stage = (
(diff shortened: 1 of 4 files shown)


davesnowdon:
Thanks — Finding 1 addressed in `f84efa2`, and an answer to your question below.

**Finding 1 (repair parse composition only weakly pinned).** You're right that `TestSweepNoteParse` re-created the production expression. Added `test_title_stripped_note_parses_archive_path_and_clears_tag`, which drives the real `sweep()` over a read-note-shaped note (no leading `# Title`, archive `path:` in the body, `influx:archive-missing` set) with **no** archive-download hook. Because the archive stage is tag-selected but hook-skipped, the only way `influx:archive-missing` clears is if `compute_clearing` saw a non-`None` archive path — which requires the parse composition (`parse_lenient` + `_doc_title`) to have extracted it from the body.

I verified it's a real guard, not a passer: temporarily dropping `_doc_title` from the caller (so `parse_lenient` has no fallback and raises on the title-stripped note → `archive_path` stays `None`) makes the test **fail**, exactly the regression you described. Reverted after confirming. Also added the archive-missing-cleared assertion to the CRLF splice test (your nice-to-have).

**Q — is it intentional that `upsert_archive_path` is a no-op when a `path:` already exists, even if the hook downloaded a different one?** Yes, intentional, and it matches the legacy behaviour byte-for-byte: the old code only inserted `if not rest.startswith("path:")`, i.e. it never overwrote an existing path either. Two things worth noting:

1. In practice the archive-download retry only fires under `influx:archive-missing`, and a note that genuinely has no stored archive won't carry a `path:` line — so insert-vs-overwrite is moot on the normal path.
2. Where the canonical op *does* differ from legacy is the direction you'd want: legacy only checked the byte immediately after the heading, so a `path:` preceded by a blank line or indentation was treated as "absent" and a **second** `path:` got inserted (duplicate). `upsert_archive_path` scans the whole `## Archive` section, so it's correctly idempotent there.

You're right that `influx:archive-missing` + an already-stored `path:` is an inconsistent state; today (and before this PR) the sweep resolves it by clearing the tag (the path is present) rather than re-downloading and overwriting. If we ever want "re-download always wins," that's a deliberate behaviour change to `upsert_archive_path` (or a separate `set_archive_path`) with its own test — happy to file it as a follow-up, but it's out of scope for this byte-preserving migration.

`make check` was green before this test-only addition (minus the known `test_run_conflict_exit_1` flake); the new test and the CRLF test pass locally, lint/format clean.

PR #255, opened 2026-07-06
refactor(repair): route repair_counters placement through canonical ops (PR 5/6)

davesnowdon:
## Context

**PR 5 of 6** in the CanonicalNote stack (finding 1 / Lithos task `9ae1086e`), on top of #251–#254. This is the deliberately **hybrid** one: `RepairCounters` keeps the domain knowledge it genuinely owns — the per-stage counter bullet grammar — and hands only the section-*placement* machinery to `influx.canonical_note`.

## What this PR does

- **`parse_repair_section`** — locates and slices the `## Repair` body via `canonical_note.extract_section_body` instead of the local `_find_repair_section_span` + `_REPAIR_HEADING_RE` / `_NEXT_HEADING_RE`. The bullet grammar (`_REPAIR_BULLET_RE`, `_QUOTED_RE`, `_collapse_to_single_line`) stays local — that's the counter domain.
- **`upsert_repair_section`** — collapses to a one-liner delegating to `canonical_note.upsert_section_text(content, REPAIR, render_repair_section(counters))`, deleting the local replace/insert logic and the `_REPAIR_PROFILE_RELEVANCE_RE` / `_REPAIR_USER_NOTES_RE` placement anchors. Byte-identical to the prior implementation (PR 1's parity harness pinned it; the whole `test_repair_counters` suite stays green unchanged).
- Kept entirely local: `RepairCounters` (+ `bump_*`/`attempts_for`), `render_repair_section`, `classify_failure`, `record_counted_failure`, `CountedFailureResult`. Net −41 lines.

## Incidental fix

Same shape as PR 4's archive-CRLF fix: the legacy read regex `^## Repair[ \t]*\n` never matched a `## Repair\r\n` heading, so a CRLF note silently parsed **zero** counters — losing terminal-flip state on the read side. `extract_section_body` is CRLF-tolerant. Added a parse test pinning it.

## Tests

- Added `TestParseRepairSectionLocation` — CRLF parse round-trip + absent-section zero-defaults.
- Retired the **last** transitional legacy-parity case (`test_upsert_repair_matches_repair_counters`): now that `upsert_repair_section` *is* the canonical op, that assertion is a self-comparison. The per-op byte-exactness it guarded is pinned directly by the canonical-op tests and `test_repair_counters`. This clears the whole transitional-parity scaffold the stack has been carrying since PR 1.

## Follow-up (final PR)

PR 6 (optional) — renderer composes via `serialize` (round-trip identity by construction), drop the dead `source_url` param from `render_note`, and converge `replace_profile_relevance_section`'s end-boundary onto the shared section-span helper (carried over from the #253 review), each with a parity test.

### DIFF
--- a/src/influx/canonical_note.py
+++ b/src/influx/canonical_note.py
@@ -639,6 +639,12 @@ def upsert_section_text(content: str, heading: str, rendered_section: str) -> st
     place (trimming leading blank lines from the following content so blank
     lines do not accumulate across re-renders); otherwise it is inserted at
     :func:`insertion_point`.
+
+    The touched section is written verbatim from *rendered_section*, i.e. with
+    the LF line endings the Influx renderers emit.  Surrounding bytes are
+    preserved as-is, so on a (non-canonical) CRLF note the replaced/inserted
+    section is LF while the rest of the note keeps its CRLF endings — the ops
+    do not round-trip CRLF for the section they own.
     """
     span = _section_span(content, heading)
     if span is not None:
--- a/src/influx/canonical_note.py
+++ b/src/influx/canonical_note.py
@@ -640,11 +640,13 @@ def upsert_section_text(content: str, heading: str, rendered_section: str) -> st
     lines do not accumulate across re-renders); otherwise it is inserted at
     :func:`insertion_point`.
 
-    The touched section is written verbatim from *rendered_section*, i.e. with
-    the LF line endings the Influx renderers emit.  Surrounding bytes are
-    preserved as-is, so on a (non-canonical) CRLF note the replaced/inserted
-    section is LF while the rest of the note keeps its CRLF endings — the ops
-    do not round-trip CRLF for the section they own.
+    The touched section — and the blank-line separators placed immediately
+    around it — are written with the LF line endings the Influx renderers
+    emit; content *below* the section (e.g. the ``## User Notes`` region) is
+    preserved byte-for-byte.  On a (non-canonical) CRLF note the section and
+    its adjacent separators therefore come out LF — a CRLF blank line just
+    above the section gains a trailing LF — while the region below keeps its
+    CRLF endings.  The ops do not round-trip CRLF for the region they own.
     """
     span = _section_span(content, heading)
     if span is not None:
--- a/src/influx/repair_counters.py
+++ b/src/influx/repair_counters.py
@@ -28,6 +28,7 @@ import re
 from dataclasses import dataclass, replace
 from typing import Any, Literal
 
+from influx.canonical_note import REPAIR, extract_section_body, upsert_section_text
 from influx.errors import ExtractionError, LCMAError
 
 __all__ = [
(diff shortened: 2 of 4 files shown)


davesnowdon:
Thanks — all three addressed in `8c6b2c5`, plus answers below.

**Finding 1 (byte-exact upsert no longer directly pinned) — fixed.** You're right that retiring the parity test left the `## Repair` placement bytes asserted only loosely, and the migration's safety rests on exactly those bytes. Replaced `TestUpsertSectionText` with exact-output assertions for the full fallback chain:
- insert before `## Profile Relevance`;
- fallback before `## User Notes`;
- fallback at EOF;
- in-place replacement (asserts the exact string *and* no `\n\n\n` accumulation).

These now pin the whole output, not just ordering.

**Finding 2 (CRLF write behaviour) — pinned + documented.** Added `test_crlf_note_touched_section_lf_surrounding_preserved` and a docstring paragraph on `upsert_section_text`. The intended contract (now explicit): the touched Influx-owned section is written **LF**, and surrounding bytes are **preserved as-is** — so on a CRLF note the `## Repair` block comes out LF while the rest (including the `## User Notes` region, byte-exact) keeps its CRLF endings. This is the same rule `upsert_archive_path` already follows for the inserted `path:` line (PR 4), so it's consistent across the canonical string ops rather than a new quirk.

**Q1 — is mixed LF/CRLF on a CRLF update allowed, or should it be normalized more deliberately?** Allowed, by design — the ops normalize the *section they own* to LF and don't round-trip CRLF for it. I deliberately did **not** touch the `rstrip("\n")`/`lstrip("\n")` trimming in this PR: that's shared machinery behind every `upsert_section_text` caller, and "make it fully CRLF-aware" is a behaviour change that could shift LF-note bytes — out of scope for a byte-preserving migration. If we ever want notes to round-trip CRLF end to end, that's a deliberate follow-up with its own parity net (and honestly the renderer only ever emits LF, so CRLF is a hand-edit/external-tool artifact, not a normal shape).

**Q2 / Finding 3 — scope the "byte-identical" claim to canonical LF notes.** Agreed, and that's the accurate scoping: byte-identical to the legacy splice **on canonical notes**. The placement now uses the anchored heading matcher (`^## Heading[ \t]*(?=\r?\n|$)`) rather than the old `^## Profile Relevance\b` / `^## User Notes\b`, so a variant like `## Profile Relevance: extra` is no longer treated as an insertion anchor. That's the same locator-tightening the whole stack has applied (PRs 2–4) — protective, and it only diverges on malformed/variant headings, never on renderer output. I've kept it as the intended contract rather than restoring the loose `\b` anchors.

`make check` was green before this test+docs-only change (minus the known `test_run_conflict_exit_1` flake); the new/updated tests pass and lint/format/pyright are clean.

davesnowdon:
Good catch — you're exactly right, and it's fixed in `8d7782a`.

**Finding 1 (CRLF contract overstated).** "Surrounding bytes preserved as-is" was wrong: in replacement mode `head = content[:start].rstrip("\n")` leaves the dangling `\r` and `separator = "\n\n"` adds an LF, so `# T\r\n\r\n` becomes `# T\r\n\r\n\n` before the LF section. I:
- Corrected the docstring to scope preservation to the content *below* the section (the `## User Notes` region, byte-for-byte), and state plainly that the touched section **and its adjacent separators** are canonical LF — explicitly calling out that a CRLF blank line just above the section gains a trailing LF.
- Changed the CRLF test to assert the **full exact output** — `"# T\r\n\r\n\n## Repair\n- tier2_attempts: 2\n\n## User Notes\r\nMINE\r\n"` — including that extra-LF artifact, instead of the loose `in` / `endswith` checks. So the boundary behaviour is now pinned precisely, not just implied.

**Q — is the extra LF intended canonicalization, or should the helper preserve the exact preceding separator bytes?** It's an incidental consequence of the LF-normalizing trim, not a designed-for outcome — but I'm accepting it rather than special-casing CRLF in `rstrip("\n")`/`lstrip("\n")`. That trim is shared by every `upsert_section_text` caller, and making it `\r`-aware is a behaviour change that could shift LF-note bytes (the real, only production shape) for zero production benefit, since the renderer only ever emits LF. So the deliberate contract is: **the section and its own separators are canonical LF; the region below is preserved byte-for-byte** — and the test now nails that exactly so it can't drift silently. If we ever want true end-to-end CRLF round-tripping, that's a separate, deliberate change with its own parity net.

`make check` was green before this docs+test-only change (minus the known `test_run_conflict_exit_1` flake); the updated test passes and lint/format/pyright are clean.

PR #258, opened 2026-07-08
feat(guardrail): extend architecture tooling — metrics, budgets, trends, per-component pages

davesnowdon:
## What

Ports the lithos [architecture-tooling uplift](https://github.com/agent-lore/lithos/commit/9d98c2fd7dd21415d5aef1e49a9b877ad5e2122c) to influx, so this repo's "diagrams as tests" system does more than draw two diagrams — it now also **explains** the codebase and **tracks whether the architecture is improving or regressing over time**. The generator kit also moves from `tests/meta/` to `tests/guardrail/` to match the sibling repos.

On top of the existing component + domain diagrams, `make diagrams` now generates, all statically (ast + grimp) and drift-checked in CI:

- **Architecture metrics** — `docs/generated/metrics.json` + `metrics.md`: coupling, cycles, size/shape, cyclomatic complexity, domain/model and test-ratio stats.
- **Hard budgets** — `[budgets]` in `docs/architecture.toml`, enforced by `test_metrics_budgets.py`. Most budgets are **tight ratchets pinned at today's value** — any regression fails CI. `max_module_lines` is the deliberate exception: a **stop-loss ceiling** (2000, set a little above today's largest module rather than pinned to it), with the *count* of oversized modules ratcheted separately via `modules_over_800_lines`. Values measured at port time: **64** cross-component edges, **3** component cycles, **3** module cycles, **12** modules over 800 lines; largest module is `sources.arxiv` at 1912 lines (stop-loss ceiling 2000).
- **Generated index** — `docs/generated/README.md`.
- **Per-component drill-down pages** — `docs/generated/components/<Component>.md` (15); component nodes in `architecture.md` are now clickable.
- **Diagram polish** — tier-skipping and foundation fan-in edges de-emphasized via `linkStyle`; domain model grouped into per-component sections with AST-derived cardinalities.
- **Trend tooling** — `scripts/metrics_history.py` + `scripts/metrics_diff.py`; a PR-only CI step posts the metrics delta vs base to the job summary.

## Notes

- **`tests/meta/` → `tests/guardrail/`** (`git mv`): the repo-specific guards `test_layering_contract.py`, `test_no_stubs.py`, and `test_v1_gate.py` move with it unchanged; the `make diagrams` target and the CI job follow the rename.
- **networkx** added to the dev group (graph SCC / condensation metrics); `uv.lock` updated (only networkx changes).
- CI drift gate hardened from `git diff --quiet` to `git status --porcelain` so newly generated, never-committed artifacts (component pages, metrics) fail the gate too.
- The kit's two **optional adapters are omitted**: influx is an MCP *client* to Lithos (its HTTP surface is FastAPI, not a decorator-registered tool surface → no tool catalog) and its file-based stores have no central store-config class to anchor a container view to.
- `docs/architecture.toml` gains `[project]`, `[component_docs]`, and `[budgets]`; `[components]`/`[tiers]` unchanged. The **domain-diagram scope is unchanged** (still `schemas` + `run` + `lcma_wiring`); `[domain].exclude_modules` was expanded to classify every other model-bearing module so the new always-on completeness gate (`test_every_model_is_scanned_or_excluded`) passes. Widen `include_modules` later to diagram more of the domain.
- The two script-import tests use an inline `# pyright: ignore[reportMissingImports]` (the scripts are added to `sys.path` at runtime).

## Test plan

- [x] Guardrail suite green — 85 tests (`pytest tests/guardrail/`), including the preserved `test_layering_contract` / `test_no_stubs` / `test_v1_gate`.
- [x] `make lint` + `make typecheck` green.
- [x] `make diagrams` is deterministic: regenerating after staging leaves `docs/generated` byte-identical.
- [x] Domain completeness + orphan guards pass; budgets hold.
- [x] CI: `diagrams` job passes on the hardened gate; PR metrics-delta step renders in the summary.

> **Note:** one pre-existing integration test, `tests/integration/test_cli_http.py::TestRunCliIntegration::test_run_conflict_exit_1`, is timing-sensitive: it passes in isolation and in CI (main is green) but times out in a local full-class run. It fails identically on pristine `origin/main` and is unrelated to this change (no `src/` code is touched here).


davesnowdon:
## Kit-level review fixes propagated from lens (30e793a)

The lens port ([lithos-lens#16](https://github.com/agent-lore/lithos-lens/pull/16)) got two reviews; since the guardrail kit is shared verbatim across all three ports, I've applied the same fixes here to keep them consistent:

- **`_common.py`** — `root_package` is now **required** (dropped the `"lithos"` fallback) so a missing/typo'd `[project]` fails loudly instead of silently scanning the wrong package; `write()` pins `newline="\n"` for byte-identical regen on Windows (the drift gate compares bytes).
- **`test_architecture_diagram.py`** — the orphan check reuses the memoized `build_import_graph()` from `_common` instead of rebuilding grimp's graph.
- **`metrics_diff.py`** — guards the empty-base case (a port's own first PR) and prints "first snapshot (no base to diff)" instead of a full `— -> value` table.
- **`_metrics_render.py`** — the `MCP tools: 0` line is dropped for repos with no tool surface, so `metrics.md` no longer advertises a 0-tool surface (regenerated below).
- **`metrics_history.py`** — the history walk now skips a malformed snapshot (not just a missing one) so one bad blob can't crash it; added `iter_snapshots` unit tests (monkeypatched `_git`) for normal / missing / malformed / empty history.
- **`Makefile`** — `metrics-diff` uses `set -e` + `trap` so it exits with `metrics_diff.py`'s status, not `rm`'s.
- **`architecture.toml [budgets]`** — added the header paragraph documenting `max_module_lines` as a **stop-loss ceiling** (already the case here: 2000, above today's largest module `sources.arxiv` at 1912), distinct from the value-pinned ratchets; oversized-module count is ratcheted separately via `modules_over_800_lines`.
- **`tests/guardrail/AGENTS.md`** — notes the fresh-port first-run test-ordering gotcha.

Lint, type-check, and `make diagrams` (deterministic) all green. One pre-existing timing flake remains unrelated to this branch — `tests/integration/test_cli_http.py::...::test_run_conflict_exit_1` times out its 10s `python -m influx run` subprocess under machine load. This branch touches **zero** files under `src/` or `tests/integration/` (`git diff origin/main...HEAD -- src/ tests/integration/` is empty), so it behaves identically to `main`; it passes on a warm CI runner.

davesnowdon:
## Code-quality comments addressed

`github-code-quality[bot]` flagged 8 cases of implicit adjacent-string concatenation inside lists in `tests/guardrail/_index.py` (the `_usage_lines` and Legend blocks) — a byproduct of wrapping long literals to the 88-char line length, which the bot warns can mask a missing comma.

Each wrapped literal is now an **explicit `+` concatenation**, matching the style already used in `render_index` in the same file. The concatenated result is identical, so `docs/generated/README.md` is byte-identical and there's no diagram drift. Lint + guardrail suite green.

PR #260, opened 2026-07-09
refactor(repair): route Tier 2 recovery through the shared Cascade (3a.3)

davesnowdon:
**PR 3 of the finding-3a stack** (repair sweep reuses the Cascade — Lithos task `62aa58f4`, design in finding `6cea71c7`). Follows #259 (3a.2). This routes the sweep's **Tier 2** recovery through the shared `Cascade.enrich` instead of the bespoke `tier2_enrich` hook, deleting the last full-text enrichment duplication in `repair_hooks`.

Scoped to **Tier 2 only** (the source-agnostic *archive re-extraction* path — reads the note's already-downloaded archive, not the create-path fetch). Tier 1 recovery is 3a.4; the `_(proposed)_` marker cleanup is 3a.5.

### Sweep runs Tier 2 through the Cascade
`_run_tier2_via_cascade` reconstructs an `Acquired` from the persisted note (doc-title, id, `## Archive` path → `archive_path`, `extracted_text=None` so the Tier-2 gate fires) and runs `enrich(acquired, score, stages={"tier2"}, counters=parse(## Repair))`. On success it inserts `## Full Text` + the `full-text` tag; on a counted failure it persists the advanced counters and flips `influx:tier2-terminal` at the cap — **`enrich` owns the counter lifecycle** (3a.1 Fork 6). A missing archive path stays a counted `parse` failure exactly as the old hook raised, so the Tier-2 cap still trips for a note that never gains an archive.

The Tier-2 and Tier-3 failure handlers now share `_apply_enrich_recovery_failure` (persist advanced counters, apply the newly-tripped terminal flag, emit the terminal-flip WARNING).

### The sweep's Tier-2 extractor
`repair_hooks.make_sweep_tier2_extractor(config)` builds the `influx.cascade.Tier2Extractor` — a source-agnostic `(Acquired) -> Tier2Result` closure over the existing `_read_archive_file` + `_extract_from_archive` helpers. `sweep()` wires it into `_build_sweep_cascade` when `hooks is None`; an absent cascade skips **both** tiers, exactly like an absent hook skips its stage.

The counted-vs-transient contract is preserved: no archive path → `parse` (**counted**, advances the counter); `archive_read` / `extract` / `min_length` → **transient** (repair-needed stays, no advance).

### Deleted the tier2 hook
`SweepHooks.tier2_enrich`, the `Tier2EnrichHook` protocol, `repair_hooks._make_tier2_enrich_hook`, the `DefaultSweepHooks` / `to_sweep_hooks` / `make_default_sweep_hooks` wiring, and the now-unused `_run_tier_retry` (both tiers run via `enrich`).

## Observability note (behaviour change, called out)
Tier-2 failures now degrade **silently** at the Cascade level — only Tier 1 / Tier 3 log a per-pass WARNING; Tier 2 just falls back to abstract-only + `influx:repair-needed`. This drops the old sweep-level `tier2_enrichment` WARNING (with `exc_info`) for per-pass failures. The counted **stage + error still survive durably** in the note's `## Repair` counters (`tier2_last_stage` / `tier2_last_error`), and the sweep still emits the `tier2_terminal_flip` WARNING at the cap — so root cause stays recoverable.

## Test migration
- **Unit** — `make_sweep_tier2_extractor` is covered directly (PDF/HTML `Tier2Result`; counted `parse` on no-path, transient `archive_read` on missing file). The sweep's Tier-2 failure test now drives a **real Cascade + fake `tier2_extractor`** and asserts the durable `## Repair` counter advance + repair-needed retention (Score 8 isolates Tier 2 from Tier 3).
- **Integration** (zero-monkeypatch) — the RSS recovery pass-2 injects one `Tier2Tier3SuccessCascade` fake driving both tiers via `sweep(cascade=...)`; the high-score-terminal test's single `SpyCascade` now proves **both** tiers are skipped. Removed the obsolete `Tier2EnrichHook` substitution test and the `tier2_enrich` rollback test (their premise — a hook mutating-then-raising — no longer exists; the extractor returns a value).

## Test plan
- [x] `make check` green — full suite; only the pre-existing `test_run_conflict_exit_1` subprocess-timeout flake (confirmed identical on clean `main`).
- [x] All migrated repair unit + integration sweep tests pass (151 in the touched files).
- [x] No `docs/generated/` drift — the `Repair → Enrich` component edge already existed (via `repair_hooks → enrich`), now fully realised via `repair → cascade`.
- [x] pyright + ruff (check + format) clean.

## Next in the stack
3a.4 routes **Tier 1** recovery through the Cascade (extend `_reconstruct_acquired` with `abstract`, regenerate `## Summary`). Then 3a.5 (cleanup + drop the `_(proposed)_` markers; possibly fold Tier 2 + Tier 3 into one `enrich` call).

### DIFF
--- a/src/influx/repair.py
+++ b/src/influx/repair.py
@@ -7,11 +7,12 @@ Tier 2, Tier 3) based on the current tag set, and rewriting every
 visited note so ``updated_at`` advances (retry-order advancement).
 
 Worker hooks (``archive_download``, ``re_extract_archive``,
-``tier2_enrich``, ``text_extraction``) are test-injectable callables
-whose real implementations ship with PRD 07.  Tier 3 recovery is no
-longer a hook: it runs through the shared
+``text_extraction``) are test-injectable callables whose real
+implementations ship with PRD 07.  Tier 2 and Tier 3 recovery are no
+longer hooks: they run through the shared
 :class:`~influx.cascade.Cascade` passed to :func:`sweep` (finding 3,
-3a.2), which the isolation tests inject the same way they inject hooks.
+3a.2 / 3a.3), which the isolation tests inject the same way they
+inject hooks.
 """
 
 from __future__ import annotations
@@ -24,7 +25,7 @@ import json
 import logging
 from collections.abc import Iterator
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Protocol
+from typing import TYPE_CHECKING, Any, Literal, Protocol
 
 from influx import metrics
 from influx.canonical_note import (
@@ -33,6 +34,7 @@ from influx.canonical_note import (
     drop_tier2_and_tier3,
     extract_section_body,
     graft_user_notes,
+    insert_full_text_section,
     insert_tier3_sections,
     parse_lenient,
     upsert_archive_path,
@@ -43,6 +45,7 @@ from influx.errors import ExtractionError, LCMAError, LithosError
 from influx.notes import merge_tags
 from influx.repair_counters import (
     CountedStage,
+    RepairCounters,
     classify_failure,
     parse_repair_section,
     record_counted_failure,
@@ -51,7 +54,7 @@ from influx.repair_counters import (
 from influx.telemetry import current_run_id, get_tracer
 
 if TYPE_CHECKING:
-    from influx.cascade import Tier
+    from influx.cascade import EnrichedSections, Tier, Tier2Extractor
     from influx.config import AppConfig
     from influx.lithos_client import LithosClient
 
@@ -66,7 +69,6 @@ __all__ = [
     "SweepHooks",
     "SweepWriteError",
     "TextExtractionHook",
-    "Tier2EnrichHook",
     "apply_abstract_only_reextraction",
     "compute_clearing",
     "select_stages",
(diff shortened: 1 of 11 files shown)


davesnowdon:
Thanks for the review — addressed all three in `75e1c0f` (test-only + docstring, no behaviour change):

**Finding 1 (Medium — Tier 2 success not proven with the real Cascade seam)** — Added `TestSweepTier2RecoveryViaCascade::test_tier2_success_recovers_full_text_and_clears_repair_needed`: drives a real Cascade via `_build_sweep_cascade(config, "ai-robotics", tier2_extractor=extractor)` where the extractor records `acquired.archive_path`. The test asserts `seen == ["arxiv/2026/04/paper.pdf"]` (a None/mis-reconstructed path fails right there), then `## Full Text` + `full-text` land and `influx:repair-needed` clears (Tier 3 waived at score 8). This proves the persisted-note → archive-path handoff the fake couldn't.

**Finding 2 (Low — terminal flip only at Cascade level)** — Added `test_tier2_counted_failure_at_cap_flips_terminal_and_logs`: note carrying `## Repair` `tier2_attempts: 2` + a fake extractor raising `ExtractionError(stage="parse")`, asserting `tier2_attempts: 3`, `influx:tier2-terminal`, and `sweep_stage == "tier2_terminal_flip"` (`tier2_attempts=3`, `stage="parse"`) — mirrors the existing Tier 3 terminal test and exercises the `repair.py` persistence/logging step.

**Finding 3 (Nit — stale `sweep()` docstring)** — The `cascade` param now documents that it drives **Tier 2 and Tier 3** recovery and that `None` skips **both**.

**Answers to your questions:**
- `sweep(cascade=...)` stays the seam for both tiers; the docstring now calls it the Tier 2/Tier 3 recovery cascade.
- Sweep-level `archive_read` transient coverage: the existing `test_tier2_counted_failure_persists_counter_via_cascade` covers the counted path; a transient (`archive_read`/`extract`) at sweep level asserts only 'repair-needed stays, no counter advance' — I kept that at the Cascade-classification level (`test_cascade.py`) rather than duplicate it, but happy to add a sweep-level one if you'd prefer it explicit.

PR #265, opened 2026-07-10
refactor(sources): route RSS scoring through Filter (2.3b)

davesnowdon:
## What

Fourth PR of the **finding-2 stack** (*finish the Source seam*) and the RSS
counterpart to #264. Routes RSS scoring through the shared
`influx.filter.Filter` seam and deletes the bespoke `_score_rss_items`
duplicate, so RSS now uses the same fetch/score/acquire seams arXiv does:
`RssSource` for fetch + acquire, `Filter` for scoring.

- `make_rss_item_provider` gains a `scorer: BatchScorer | None` test seam
  (symmetric with arXiv). `None` installs the production default
  `make_default_batch_scorer(config)` — the same source-agnostic LLM filter
  scorer arXiv uses. `make_item_provider` threads it through as `rss_scorer`.
- The provider flattens every configured feed into one candidate list
  (`RssSource.fetch_candidates`), scores it via `Filter.score(source="rss")`,
  and binds each survivor via `RssSource.acquire`.
- `_score_rss_items` deleted.

## ⚠️ Production behaviour changes (please review before merge)

This is the PR we agreed to look at before merging, because it changes the
production filter call pattern and failure handling:

1. **Fewer, fuller LLM filter calls.** `Filter` chunks the *flattened*
   candidate list by `filter.batch_size`, so batches pack **across feed
   boundaries** instead of restarting per feed. A multi-feed profile makes
   fewer filter calls (lower LLM cost). Scoring is otherwise identical — same
   prompt, same threshold gate, same `articles_filtered{drop,pass}` metrics
   (now emitted by `Filter` rather than the provider loop).

2. **Filter failure skips the failed batch, keeps clean ones.** A
   `FilterScorerError` skips **only the batch it occurred in** (FR-FLT-6 /
   spec §7.1: *"failed filter batches are skipped"*) and records one
   `filter_error` per failed batch; batches that already scored cleanly are
   still ingested. Failed-batch items are dropped **without** an
   `articles_filtered{drop}` tick, preserving the 2.3a decision that *"a
   failed filter call is not a per-item rejection"*. This matches the old
   per-feed `_score_rss_items` resilience (a transient failure never discarded
   already-scored items). _(Corrected after review — see §Update below.)_

3. **Per-item + per-feed INFO logs removed.** The per-item `article inspected`
   logs and the per-feed `rss filter completed` / `rss feed completed` logs
   lived in the deleted provider loop and are gone. The aggregate
   `rss source completed profile=.. feeds=.. fetched=.. accepted=..` summary
   log is kept (and now also reports the fetched count). No test asserted the
   removed logs.

## Update (review fix, commit a56c700)

The reviewer correctly caught that the *shared* `Filter.score` did **not**
match the "skip the failed batch" description above: it did `return []` on the
first `FilterScorerError`, discarding candidates from earlier batches that had
already scored cleanly (whole-call-drop). Since RSS/arXiv now make one
flattened `Filter.score` call, a single transient failure on a later batch
would drop the whole run.

Fixed in `Filter.score`: record `filter_error` and `continue` instead of
`return []`, tracking the failed batch's item ids so they're dropped without a
per-item drop tick. This aligns the code with the spec's per-batch language and
with §2 above. It affects **both** callers of the shared seam — RSS (this PR)
and arXiv (#264) — and only diverges from the old behaviour once a run spans
more than one batch (single-batch runs still yield `[]` on failure).

## Tests

- **Deleted 3** tests that drove the now-removed `_score_rss_items` directly
  (chunk-by-batch-size / skip-failed-chunk / retry-after-429) — that logic now
  lives in and is tested by `influx.filter`.
- **Migrated** the remaining RSS scorer seams onto injected `BatchScorer`s:
  `test_rss_fetcher`, `test_telemetry_wiring`, and the two dedup integration
  tests (`test_fetch_dedup`, `test_pre_acquire_dedup`).
- **Added** (review fix) `test_filter.py::test_score_keeps_successful_batches_when_a_later_batch_fails`
  (the authoritative multi-batch pin, inherited by both sources) and
  `test_rss_fetcher.py::TestFilterScorerErrorSkipsFailedBatch::test_earlier_batch_survives_later_batch_failure`
  (multi-item RSS provider, `batch_size=1`, `[Fail C]` batch fails → `Keep A` /
  `Keep B` survive). Renamed the RSS failure class to `…SkipsFailedBatch`.
- `make check` green locally: **2752 passed**, lint + pyright clean.

## Metrics

`docs/generated/metrics.*` regenerated — Sources drops **4588 → 4418** lines
(the deleted provider loop); Filter grows **550 → 569** (the review fix). The
diff also reconciles a **pre-existing staleness** in the committed metrics for
`arxiv.py` (1902 → 1885): byte-identical to `main`; #264 changed it but did not
co-commit the regenerated metrics.

## Scope

Refs Lithos task `989db18f` (finding 2, scope 2.3b). Follows #264 (2.3a).
Inbox scoring stays deliberately divergent and is handled in 2.4 (inbox ↔
Source decision + cleanup).


davesnowdon:
Thanks — this was a real correctness bug, not just a wording mismatch. You're right that `Filter.score` did `return []` on the first `FilterScorerError`, discarding candidates from earlier batches that had already scored cleanly. Because RSS (and arXiv since #264) now make one flattened `Filter.score` call, a transient failure on a later batch would drop the whole run.

I went with **"skip only the failed batch, keep the clean ones"** as the contract, because:
- The spec says *"failed filter **batches** are skipped"* (§7.1 / FR-FLT-6) — per batch, not per call.
- It's what the old per-feed `_score_rss_items` did (failed chunk → `record_filter_error()` + `continue`), so it removes the RSS resilience regression this PR would otherwise introduce.
- It's strictly better in production: a one-batch LLM hiccup shouldn't throw away 100 already-scored candidates.

Fix in `a56c700` (`Filter.score`): record `filter_error` and `continue` instead of `return []`, tracking the failed batch's item ids so they're dropped **without** an `articles_filtered{drop}` tick — preserving the 2.3a decision that a failed filter call is not a per-item rejection. This is in the shared seam, so it fixes **arXiv too**; single-batch runs still yield `[]` on failure, so behaviour only diverges once a run spans >1 batch. `filter_error` is now recorded once per failed batch (was once per call), matching old RSS; the ledger only gates on `> 0`.

Tests added for exactly the case you flagged:
- `test_filter.py::test_score_keeps_successful_batches_when_a_later_batch_fails` — 4 candidates, `batch_size=2`, the second batch raises → the first batch's items survive, one `filter_error` recorded. This is the authoritative multi-batch pin and is inherited by both sources.
- `test_rss_fetcher.py::TestFilterScorerErrorSkipsFailedBatch::test_earlier_batch_survives_later_batch_failure` — multi-item RSS provider (`batch_size=1`), the `[Fail C]` batch fails → `Keep A` / `Keep B` still bound. (Renamed the class from `…DropsWholeBatch` to `…SkipsFailedBatch`.)

Updated the PR description accordingly. `make check` green (2752 passed).

PR #267, opened 2026-07-11
refactor(sources): unify arXiv+RSS on one scoring path, break Sources↔Filter cycle (2.4)

davesnowdon:
## What

**Final PR of the finding-2 stack** (*finish the Source seam*). Unifies arXiv
onto the same source-agnostic `BatchScorer` path RSS already uses, deletes the
duplicate arXiv scorer, and breaks the Sources↔Filter cycle. You picked **"unify
on one scorer"** for the scope.

## One scoring path

- `make_arxiv_item_provider` now takes `scorer: BatchScorer | None` (symmetric
  with RSS); `None` installs `make_default_batch_scorer(config)`.
- Deleted `make_default_arxiv_filter_scorer` (`filter.py`), the two adapters
  (`_batch_scorer_from_item_scorer` / `_batch_scorer_from_arxiv_filter_scorer`),
  and the `ArxivScorer` / `ArxivFilterScorer` / `ArxivScoreResult` types.
- `create_app` / `make_item_provider` drop `arxiv_filter_scorer`; arXiv defaults
  to `make_default_batch_scorer` inside the provider (like RSS).
- **Behaviour-preserving**: arXiv's `Candidate.item_id == arxiv_id`, so the
  source-agnostic scorer sends the LLM an identical request (same id, title,
  abstract). The `influx.filter` span now wraps the scoring call whenever a
  scorer is wired (production default OR injected), vs only the old
  batched-production path — no test needed weakening for this.

## Break the Sources↔Filter cycle (2 parts — please sanity-check part 2)

1. Removing the `filter → sources.arxiv` import (the finding's specific evidence)
   takes `influx.filter` out of **every module cycle**; `filter.py` now imports
   only Foundation + the `influx.source` contract types.
2. ⚠️ **Component-classification decision:** `influx.source` (the shared
   `Candidate` / `ScoredCandidate` / `Source` Protocol contract) was grouped in
   the **Sources** component, so `filter.py`'s legitimate
   `from influx.source import Candidate` still counted as a Filter→Sources
   *component* edge. `source.py` is Foundation-pure (imports only `config` +
   `coordinator`), so I moved it from the Sources component to **Common**
   (Foundation) — the standard "extract the shared contract to break the cycle"
   move. Result:

   | | before | after |
   |---|---|---|
   | module cycles | `filter ↔ sources ↔ sources.arxiv ↔ sources.rss` | `sources ↔ sources.arxiv ↔ sources.rss` (Filter gone) |
   | component cycles | `Enrich ↔ Filter ↔ Repair ↔ Sources ↔ Storage` | `Enrich ↔ Repair ↔ Sources ↔ Storage` (Filter gone) |

   import-linter layering contracts (`Foundation must not import Core`, `Core
   must not import Entrypoints`) both stay **KEPT**; all 85 guardrail tests pass.
   The remaining component cycles belong to other findings (Enrich/Repair/Storage
   tangle → finding 3; HttpApi↔Orchestration → finding 4).

## Inbox stays divergent (documented, no code change)

Inbox already scores via `make_default_batch_scorer` directly — a per-Profile
fan-out over *one* submitted item, deliberately not through `Filter.score` so it
keeps its own inbox metrics. It isn't a bulk-fetch Source, so it stays a separate
fan-out rather than being forced onto the seam. The CONTEXT.md `Source` entry
drops its `_(proposed)_` marker and records this.

## Tests

- Migrated the arXiv scorer call sites across **8 files** onto injected
  `BatchScorer`s (same pattern as the RSS 2.3b migration). No assertion weakened;
  each per-item helper's original scoring intent preserved.
- **Deleted** `test_arxiv_filter_adapters.py` (tested the removed adapters).
- `make check` green: **2745 passed**, lint + pyright + guardrails clean. The one
  failure locally is the pre-existing sandbox flake `test_run_conflict_exit_1`
  (a `python -m influx run` subprocess that trips the test's 10s startup timeout
  in the sandbox; passes in CI on every prior PR).

## Scope

Refs Lithos task `989db18f` (finding 2, scope 2.4). Follows #265 (2.3b).
**Completes the finding-2 stack** (2.1 RSS fetch → 2.2 note-builder → 2.3a/2.3b
scoring → 2.4 unify + cycle break).


davesnowdon:
Thanks — added both, in `0f08111`:

**Payload assertion (the Low finding).** `test_default_scorer_below_full_text_yields_abstract_only` now decodes the captured `guarded_post_json_fetch` body and asserts the `## CANDIDATES` JSON is exactly `[{"id": <arxiv_id>, "title": ..., "abstract": ...}]`. Since the mock response was independent of the request, this is what actually pins `Candidate.item_id == arxiv_id` and the title/abstract mapping against a future regression.

**No-scorer provider coverage (your nice-to-have).** Good catch — that case *was* covered in `test_arxiv_filter_adapters.py`, which this PR deletes, so the coverage went with it. Restored as `test_telemetry_wiring.py::TestInfluxFilterSpan::test_no_filter_span_and_zero_items_when_no_scorer`: `make_arxiv_item_provider(config_without_models_filter)` with no scorer → fetches candidates but yields zero items and emits no `influx.filter` span.

**On `arxiv_scorer` naming (your question).** Intentional. `make_item_provider` already exposes `arxiv_scorer` / `rss_scorer`, and both `make_arxiv_item_provider` / `make_rss_item_provider` take `scorer` — all now the same `BatchScorer` type. Keeping `arxiv_scorer` (not `arxiv_batch_scorer`) keeps it symmetric with `rss_scorer`; renaming one without the other would desync the pair. Happy to rename both to `*_batch_scorer` in a later sweep if you'd prefer, but I left the pair consistent for now.

Test-only changes; no src touched. `make check` still green apart from the known `test_run_conflict_exit_1` sandbox flake.

PR #269, opened 2026-07-11
refactor(repair): preserve ## Repair counters across multi-profile merge (3a.4)

davesnowdon:
## What

Finishes the **create-path half of finding-3a's RepairCounters wiring** (scope 3a.4). Only the repair sweep writes a note's `## Repair` section; the create path enriches with a zero counter and never emits one. So when an existing — possibly **tier-terminal** — note is re-ingested by another profile, `LithosClient._retry_version_conflict` merged the fresh create-path content (no `## Repair`) over the existing note and **dropped the accumulated counters**, resetting a capped tier to zero. The next sweep would then re-attempt that tier from scratch, forever — the terminal cap was never respected on re-ingest.

## Fix

Carry the existing note's `## Repair` section forward verbatim into the merged content, mirroring how `graft_user_notes` / `_merge_profile_relevance_in_content` already preserve existing sections at the merge.

- New `canonical_note.carry_forward_section(source, target, heading)` — a generic section-copy op (`extract_section_body` + `upsert_section_text`). It carries the section *body* as text (no `RepairCounters` parse/re-render round-trip, so unknown/hand-added bullets survive), but — like every canonical section op — normalises the section it writes to LF boundaries. So it's **semantic** preservation of the counter content, **not** byte-exact: a hand-edited `## Repair` keeps its bullets but loses trailing blank lines / CRLF within the section. _(Wording corrected after review; exact-output + normalisation tests added.)_
- Kept in `canonical_note` (not `repair_counters`) deliberately: `LithosClient` (LithosBridge) importing `repair_counters` (Repair) would add a **LithosBridge↔Repair component cycle** (`repair.py` already imports `lithos_client`). The verbatim copy needs only the section ops, no bullet grammar.

A capped note's counters now survive re-ingest. Note the `influx:tierN-terminal` **tag** is Influx-owned and intentionally *replaced* by `notes.merge_tags` (the fresh write re-derives its own `influx:*` tags), so it's the preserved **counters**, not the tag, that carry the cap — the next sweep re-derives the terminal tag from `attempts ≥ cap`. The contract test documents this.

## ⚠️ Approach decision (Option B vs A) — worth a look

The `cascade.py` docstring anticipated the create-path merge **passing counters into `enrich`** (Option A). But `enrich` runs in the Source at *Acquire* time — before the merge target is read — so Option A would thread the existing counters through `acquire closure → Source.acquire → build_*_note_item → enrich` across **all three sources**, i.e. through the very Source seam **finding-3b** reworks.

I chose **Option B** (preserve the section at the merge): localized to the one function that already reads the existing note, achieves the same goal ("re-ingest respects caps"), and leaves the Source seam untouched for 3b. Trade-off: the fresh enrich still spends one (bounded, rare) re-attempt on re-ingest, and a re-ingested note may briefly carry `influx:repair-needed` until the next sweep re-terminalizes it from the preserved counters — no worse than today and self-healing. Updated the `cascade.py` docstrings + `CONTEXT.md` RepairCounters entry to describe the actual mechanism (dropped its `_(proposed)_` marker).

## Tests

- `TestCarryForwardSection` — the canonical section-copy op (copy / no-op / replace-existing).
- `test_version_conflict_preserves_repair_section` — contract test vs `FakeLithosServer`: a tier-terminal note re-ingested by a second profile keeps its `## Repair` counters + user notes; fresh content still merges.
- `make check` green: **2752 passed**, lint + pyright + guardrails clean. Only the known `test_run_conflict_exit_1` sandbox flake fails locally (subprocess-startup timeout; passes in CI).

## Scope

Refs Lithos task `62aa58f4` (finding 3, scope **3a.4**). **Tier 1 recovery** — the remaining 3a-labelled item — is folded into **3b** (task `7103dd79`), since re-running Tier 1 needs the source *abstract*, which only 3b's Source-seam re-acquire provides. 3b is next (now unblocked by finding 2).

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -43,7 +43,7 @@ Produces a **CanonicalNote** from an Acquired plus EnrichedSections plus the sco
 An Influx-authored Markdown note: typed frontmatter, fixed section order (`## Archive`, `## Summary`, `## Full Text`, `## Claims`, `## Datasets & Benchmarks`, `## Builds On`, `## Open Questions`, `## Profile Relevance`, `## User Notes`), and stable tag conventions. `## User Notes` is preserved byte-exactly across rewrites.
 
 **RepairCounters**:
-Per-tier attempt counter persisted in the note's `## Repair` section. Read on tier entry (skip if `tier{N}-terminal` is set), advanced on counted (parse/validate) failures, never advanced on transient failures. Reaches the cap → adds `influx:tier{N}-terminal`. _(proposed as a module shared between the create path and the repair sweep)_
+Per-tier attempt counter persisted in the note's `## Repair` section. Read on tier entry (skip if `tier{N}-terminal` is set), advanced on counted (parse/validate) failures, never advanced on transient failures. Reaches the cap → adds `influx:tier{N}-terminal`. The repair sweep owns the full lifecycle (read → advance → terminal) via `Cascade.enrich`; the create path enriches with a zero counter but preserves an existing note's `## Repair` section across a multi-profile re-ingest, so caps survive.
 
 ### Domain — execution
 
--- a/docs/generated/components/LithosBridge.md
+++ b/docs/generated/components/LithosBridge.md
@@ -40,7 +40,7 @@ Renders CanonicalNotes and is the write-mostly MCP/SSE client to Lithos (WriteRe
 - def `drop_tier2` — Remove the ``## Full Text`` (Tier 2) section, keeping Tier 1/Tier 3.
 - def `drop_tier2_and_tier3` — Remove ``## Full Text`` and all Tier 3 sections, keeping Tier 1.
 - def `extract_section_body` — Return the trailing-stripped body of *heading*, or ``""`` if absent.
-- def `carry_forward_section` — Copy *heading*'s section from *source* into *target* verbatim.
+- def `carry_forward_section` — Copy *heading*'s section from *source* into *target*.
 - def `upsert_archive_path` — Set the ``## Archive`` ``path:`` line, idempotently.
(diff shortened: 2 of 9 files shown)


davesnowdon:
You're right — "byte-exact/verbatim" overclaimed. `extract_section_body` rstrips the body and `upsert_section_text` re-renders LF boundaries (the normalisation every canonical section op applies to the section it writes), so it's semantic preservation of the counter content, not a byte copy. I kept the helper consistent with the `upsert_section_text` family rather than introducing a byte-exact span-copy that would deviate from it — and corrected the wording to match.

Fixed in `c3d68a3`:
- Reworded the `carry_forward_section` docstring + the PR description: the body is carried as text (no `RepairCounters` parse/re-render round-trip, so unknown/hand-added bullets survive), but section boundaries are normalised to LF and trailing whitespace within the section is dropped.
- Added the exact-output test you asked for: `test_exact_output_for_canonical_lf_content` pins the full merged string for canonical LF content; `test_normalises_trailing_blank_lines` documents the normalisation (source `## Repair` extra blank lines collapse); `test_inserts_before_profile_relevance` pins SECTION_ORDER placement.

Byte-exact preservation isn't a goal for `## Repair` (it's a machine-written counter section; the bullets are what carry the caps), so semantic preservation is sufficient — I went with the wording/test fix rather than a new span-based op. `make check` still green.

davesnowdon:
Fixed in `e524452` — swapped the two remaining "verbatim" mentions (the `_retry_version_conflict` call-site comment and the `TestCarryForwardSection` class docstring) for "as text" / "LF-normalised", matching the helper docstring. Comments only. Thanks for the thorough pass.

PR #278, opened 2026-07-29
fix(repair): waive clearing conditions for per-stage terminal tags

davesnowdon:
## Problem

Dave reported being repeatedly recommended the same agent-protocol articles. The filed diagnosis blamed duplicate ingestion; it isn't. The real cause is **write amplification in the repair sweep**.

`select_stages` honours all four `influx:<stage>-terminal` tags when deciding which stages to *run* (`repair.py:342-383`). `compute_clearing` honoured only `influx:text-terminal` when deciding whether to clear `influx:repair-needed`. So a note whose Tier 2 / Tier 3 / archive stage had exhausted `REPAIR_COUNTED_CAP` kept `influx:repair-needed` forever — even though `select_stages` would never run those stages again.

The sweep deliberately rewrites every visited note so `updated_at` advances (AC-X-8 retry-order advancement, which prevents candidate starvation when the backlog exceeds `max_items_per_run`). Combined with a note that never leaves the sweep set, that means **one spurious version per sweep, forever**.

Lithos retrieval weights recency, so these notes were permanently the freshest thing in the KB.

## Evidence from production

- **24 of 25** notes carrying `influx:repair-needed` were rewritten on 2026-07-29, out of 2225 `ingested-by:influx` notes.
- Note `0a3a8e2b` ("A Mike's-Eye View of ARC's Research") — carries **both** `influx:tier2-terminal` and `influx:tier3-terminal`, so nothing further is achievable — sits at **version 125**, created 2026-06-10 (~2.5 rewrites/day for 49 days).
- Rewrite timestamps cluster by profile (03:06 `ai-agents`, 04:18 `knowledge-systems`, 04:55 `ai-foundations`, 05:35 `ai-coding`, 06:15 `retro-computing`) = one rewrite per note per profile sweep.
- Four of the 24 are agent-protocol articles, which is why that cluster dominated every "agent protocols" query and read as duplicate ingestion.

## Change

Each clearing condition is now waived by its own terminal tag, mirroring the existing `influx:text-terminal` exemption (AC-X-7):

| Condition | Waived by |
|---|---|
| (a) archive path stored | `influx:archive-terminal` |
| (c) Tier 2 satisfied | `influx:tier2-terminal` |
| (d) Tier 3 satisfied | `influx:tier3-terminal` |

`clear_archive_missing` is deliberately **unchanged** — `influx:archive-missing` states a fact about the note (no path is stored), not the reachability of a stage.

Side benefit: this makes the `unsupported` archive-policy escape hatch effective. Previously even an operator-declared `influx:archive-terminal` could not release a note from the sweep.

## Re-arm contract (corrected in review)

Making terminal stages release notes exposed that the documented operator re-arm workflow was wrong, on three doc surfaces. **The persisted counter is authoritative, not the tag**: `Cascade.enrich` checks `tier{N}_attempts >= REPAIR_COUNTED_CAP` before calling the extractor and re-emits the terminal tag from the counter alone.

Re-arming Tier 2/3 therefore takes three edits — reset the stage counter below the cap, remove the terminal tag, re-add `influx:repair-needed` — now documented in `SPECIFICATION.md` §11.1, `CONTEXT.md`, and `runbook.md` §6, and pinned by `TestSweepTier2ReArmContract` in both directions. No runtime change: the counter was already authoritative, the docs just claimed otherwise.

Also corrected: candidate selection is profile-qualified (`influx:repair-needed` **and** `profile:<profile>`), not repair-needed alone.

## What I deliberately did NOT do

My first attempt was a diff-guard skipping the `lithos_write` when a sweep pass produced no change. **That was wrong** and is not in this PR. It breaks AC-X-8: the sweep selects candidates `order_by="updated_at", order="asc"`, so the rewrite is what rotates the cursor. Skipping it would leave unchanged notes at the front of the queue forever and starve the rest of the backlog. It failed 29 existing tests, including `test_no_progress_still_rewrites` — which encodes the invariant explicitly. The correct fix is to stop notes being immortal sweep candidates, not to stop writing them.

## Known remaining case (not fixed here)

Notes whose archive download fails **transiently** forever — e.g. a URL that always 403s — never reach a counted cap, so they stay in the sweep legitimately per the transient-retry design and keep churning. Dave's two articles (`0cff8956`, `3b1500c5`) are in this class: `influx:archive-missing` + `influx:text-terminal`, `archive_attempts: 0`. Their `source:ai-agents-briefing` has no registered reacquirer, so `_reacquirer_for_source` returns `None` and `_raise_unsupported_source` raises `stage="unsupported_source"`, which the archive path leaves transient by design.

Two options, both beyond this PR's scope:
1. Operator declares the domains `unsupported` in archive policy — **does not help this class**, because `download_archive` (where policy is consulted) is never reached.
2. Classify `unsupported_source` as counted on the archive path, so a source with no resolver goes terminal instead of retrying forever. Composes with this PR's `archive_ok` waiver. Needs its own issue — it reverses a documented intent.

## Test plan

- [x] `make check` green — 2788 passed, ruff + pyright clean
- [x] 9 cases in `tests/unit/test_repair_clearing_rules.py::TestPerStageTerminalExemptions` — positive waivers for archive/Tier 2/Tier 3, a regression case for note `0a3a8e2b`, and negative controls (terminal tiers must not bypass the text condition; each tier's waiver is scoped to its own condition; `archive-missing` stays factual)
- [x] Sweep-level convergence: Tier 2 and Tier 3 cap-flips assert `influx:repair-needed` removal survives into the persisted write; new archive test proves `archive-terminal` releases an otherwise-complete note while retaining `archive-missing`
- [x] `TestSweepTier2ReArmContract` — capped counter with tag removed invokes the extractor **zero** times and re-terminates; counter reset below cap invokes it once and recovers full-text
- [x] `test_tier2_counted_failure_at_cap_flips_terminal_and_logs` updated — it asserted `influx:repair-needed` *stays* after a cap flip, which is precisely the bug
- [x] `docs/generated/` regenerated
- [ ] Post-deploy: confirm the 24 notes stop accruing versions; those with terminal tags should drop out of the sweep on their next pass

## Note for review

The affected notes keep a misleadingly recent `updated_at` after this lands — they will simply stop advancing. Correcting the existing timestamps would itself require a write, so that is better done Lithos-side if it matters.

https://claude.ai/code/session_01QCT69UAC62JFJDX7AcHSAb

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -137,7 +137,7 @@ _Avoid_: probes (one input to Health), readiness (one output of Health).
 >
 > **Dev:** "If **Tier 2** extraction fails three times for one Acquired, what happens on the next scheduled Run?"
 >
-> **Domain expert:** "**RepairCounters** sees `tier2_attempts >= cap`, the **Cascade** skips Tier 2, the **Renderer** emits the note with `influx:tier2-terminal`, and the note drops out of the sweep set. To re-enable that stage on that note an operator removes the terminal tag *and* re-adds `influx:repair-needed` — sweeps select on that tag alone, so dropping the terminal tag by itself re-arms nothing."
+> **Domain expert:** "**RepairCounters** sees `tier2_attempts >= cap`, the **Cascade** skips Tier 2, the **Renderer** emits the note with `influx:tier2-terminal`, and the note drops out of the sweep set. Re-arming that stage takes three edits: reset the counter below the cap, remove the terminal tag, re-add `influx:repair-needed`. The counter is the authoritative gate — the **Cascade** checks it before calling the extractor, so removing the tag alone gets you a sweep that attempts nothing and re-terminates."
 
 ## Flagged ambiguities
 
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -137,7 +137,7 @@ _Avoid_: probes (one input to Health), readiness (one output of Health).
 >
 > **Dev:** "If **Tier 2** extraction fails three times for one Acquired, what happens on the next scheduled Run?"
 >
-> **Domain expert:** "**RepairCounters** sees `tier2_attempts >= cap`, the **Cascade** skips Tier 2, the **Renderer** emits the note with `influx:tier2-terminal`. Operator removes the tag manually to re-enable that stage on that note."
+> **Domain expert:** "**RepairCounters** sees `tier2_attempts >= cap`, the **Cascade** skips Tier 2, the **Renderer** emits the note with `influx:tier2-terminal`, and the note drops out of the sweep set. To re-enable that stage on that note an operator removes the terminal tag *and* re-adds `influx:repair-needed` — sweeps select on that tag alone, so dropping the terminal tag by itself re-arms nothing."
 
 ## Flagged ambiguities
 
--- a/docs/SPECIFICATION.md
+++ b/docs/SPECIFICATION.md
@@ -577,7 +577,11 @@ Tier 2 and Tier 3 hook failures are partitioned into transient (HTTP, transport,
 
 When `tier{N}_attempts` reaches the cap (currently 3), `influx:tier{N}-terminal` is added to the note's tags and that stage is skipped on subsequent sweeps. Transient failures never advance the counter, so flaky network conditions do not burn the cap.
 
-To re-enable a stage on a single note, an operator removes the `influx:tier{N}-terminal` tag (and optionally clears the `## Repair` section) — the next sweep retries from a clean slate. This mirrors the existing `influx:text-terminal` escape hatch.
+To re-enable a stage on a single note, an operator removes the `influx:tier{N}-terminal` tag (and optionally clears the `## Repair` section) **and re-adds `influx:repair-needed`** — the next sweep then retries from a clean slate.
+
+Both steps are required. Sweeps select candidates by the `influx:repair-needed` tag alone, and a terminal flip waives the clearing condition that tag gated (§5.3), so the note drops out of the sweep set when it goes terminal. Removing the terminal tag on its own leaves a note no sweep will ever select again.
+
+This mirrors the existing `influx:text-terminal` escape hatch, which has always behaved this way: `audit_invalid_source.reconstruct_tags` drops `influx:text-terminal` **and** re-adds `influx:repair-needed` for exactly this reason, and is the reference implementation for a re-arm.
 
 ---
 
--- a/docs/SPECIFICATION.md
+++ b/docs/SPECIFICATION.md
@@ -577,11 +577,32 @@ Tier 2 and Tier 3 hook failures are partitioned into transient (HTTP, transport,
 
 When `tier{N}_attempts` reaches the cap (currently 3), `influx:tier{N}-terminal` is added to the note's tags and that stage is skipped on subsequent sweeps. Transient failures never advance the counter, so flaky network conditions do not burn the cap.
 
-To re-enable a stage on a single note, an operator removes the `influx:tier{N}-terminal` tag (and optionally clears the `## Repair` section) **and re-adds `influx:repair-needed`** — the next sweep then retries from a clean slate.
+#### Re-arming a terminal stage
 
-Both steps are required. Sweeps select candidates by the `influx:repair-needed` tag alone, and a terminal flip waives the clearing condition that tag gated (§5.3), so the note drops out of the sweep set when it goes terminal. Removing the terminal tag on its own leaves a note no sweep will ever select again.
+**The persisted counter is authoritative, not the tag.** Two different mechanisms read the terminal state, and a re-arm has to satisfy both:
 
-This mirrors the existing `influx:text-terminal` escape hatch, which has always behaved this way: `audit_invalid_source.reconstruct_tags` drops `influx:text-terminal` **and** re-adds `influx:repair-needed` for exactly this reason, and is the reference implementation for a re-arm.
+| State | Read by | Effect |
+| ----- | ------- | ------ |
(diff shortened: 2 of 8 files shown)


davesnowdon:
## Review follow-up — both points addressed in 49948ca

**P2 (terminal re-arm) — fixed, with one correction to the framing.**

The substance is right and it needed fixing before merge. `docs/SPECIFICATION.md:580` and `CONTEXT.md:140` both told operators that removing `influx:tier{N}-terminal` is sufficient to re-arm a note. It isn't: sweeps select on `influx:repair-needed` alone, and the terminal flip waives the condition that tag gated, so the note leaves the sweep set. Both docs now state the required two steps.

Where I'd push back is on "regression". The spec cites the `influx:text-terminal` escape hatch as the precedent this mirrors — and that hatch has *always* worked this way. `audit_invalid_source.reconstruct_tags` step 4 says so explicitly:

> Add `influx:repair-needed` so the next sweep actually picks the note up (**the in-band terminal flip cleared it**; we re-arm it here because the metadata is now repaired).

So the documented one-step workflow was already wrong for the case it named as precedent. This PR extends the same established semantics to three more tags, which widens the inaccuracy from one tag to four — worth fixing either way, but it isn't new behaviour diverging from a previously-correct contract. Both docs now point at `reconstruct_tags` as the reference implementation.

I did **not** add a generic re-arm helper. There is no CLI surface for tier/archive-terminal re-arm today (operators edit tags in Lithos directly), and adding an uncalled helper would be dead code. Happy to file a follow-up for an operator-facing re-arm command if that's wanted — say the word.

**P3 (archive-terminal negative control) — fixed, good catch.**

Exactly right: `test_archive_terminal_does_not_clear_archive_missing` never supplied `influx:archive-terminal`, so it asserted nothing about the coupling, and the positive case didn't check `clear_archive_missing`. An incorrect coupling would have passed both. The negative control now supplies the terminal tag, and both tests assert both fields, so the two decisions are pinned as independent in each direction.

**Validation:** `make check` green — 2782 passed, ruff + pyright clean, `docs/generated/` regenerated. No behaviour change in this commit.

davesnowdon:
## All three findings addressed in b631cc8

`make check` green — 2788 passed, ruff + pyright clean.

### Finding 1 — confirmed, and worse than the review states

You're right, and this was a real blocker. I verified `cascade.py:270`: the capped-counter branch appends `influx:tier2-terminal` and never reaches `_run_tier2`. So the workflow I documented one commit earlier produced a sweep that selected the note, attempted nothing, re-terminated, and dropped it back out — silently, exactly as you describe.

**Resolved in favour of the counter as authoritative** (no runtime change — that's already how it behaves; only the docs claimed otherwise). Re-arming Tier 2/3 is now documented as three required edits: reset the stage counter below the cap, remove the terminal tag, re-add `influx:repair-needed`.

Updated consistently across all three surfaces. The runbook was furthest behind, as you noted — it claimed tag removal alone sufficed *and* that the next sweep "retries from 0", both wrong. It now also records that archive re-arm buys exactly one attempt before a capped counter re-terminates it, and that `text-terminal` has no counter.

### Finding 2 — added, including the coupling gaps

- Tier 3 cap-flip now asserts `influx:repair-needed` removal (that fixture is otherwise complete, so the waiver is the last outstanding condition).
- New `test_archive_terminal_releases_otherwise_complete_note` — the existing archive fixture carries no `text:*` tag, so condition (b) failed independently and it could never have shown release. The new one satisfies text and both tiers, and pins that `influx:archive-missing` survives the release.
- New `TestSweepTier2ReArmContract` — capped counter with the tag removed invokes the extractor **zero** times and re-terminates; counter reset below the cap invokes it **once** and recovers full-text. This is the test that would have caught Finding 1.
- Cross-tier negative controls: `tier2-terminal` alone must not waive Tier 3, `tier3-terminal` alone must not waive Tier 2, `archive-terminal` alone must not waive either.

### Finding 3 — correct, fixed

"That tag" resolved to `influx:archive-terminal`, reversing the meaning. Now names `influx:archive-missing` explicitly.

### Answers

**1. Terminal tag or counter authoritative?** The counter, for execution. The tag does two narrower jobs: gates stage selection in `select_stages`, and waives the §5.3 clearing condition. I chose to document rather than change behaviour because resetting the counter in code on tag-removal would mean inferring operator intent from "capped counter but no tag", and the sweep can't distinguish that from a state it produced itself mid-pass.

**2. Reset only the relevant counter?** Yes — both docs now say to reset only the stage being re-armed, explicitly because deleting the whole `## Repair` block discards the other stages' attempt history and last-error diagnostics. The old "optionally delete the `## Repair` block" advice was actively harmful.

**3. Operator command preferable?** Yes, clearly — three coordinated edits across tags and Markdown body is a poor operator interface and the docs are now the only thing preventing the mistake. I have not added one here: there's no existing operator-command surface for tier/archive re-arm to extend (`audit_invalid_source` is source-invalid-specific), so it's new scope rather than a fix. Happy to file it as a follow-up.

**4. Are `profile:*` tags guaranteed?** No, and my "repair-needed alone" wording was wrong — the query is `tags=["influx:repair-needed", f"profile:{profile}"]`. A note that has lost its `profile:*` tag is unreachable by any sweep whatever its repair tags say. Corrected in the spec and runbook.

### On the test count

Right on both counts — the class had six methods, not seven, and the seventh case I was counting was the modified Tier 2 sweep assertion. PR description corrected; the class now has nine.

### Still out of scope, flagged for the record

The transient-forever class is untouched and is what actually affects Dave's two articles. Their `source:ai-agents-briefing` has no registered reacquirer, so `_raise_unsupported_source` raises `stage="unsupported_source"`, which the archive path leaves transient by design — they never reach a cap, so no waiver in this PR can release them. Worth noting the archive-policy `unsupported` route also cannot help, since `download_archive` (where policy is read) is never reached. Tracked in the PR description as needing its own issue.

Issue #248, opened 2026-06-28
Repair sweep has no archive-download resolver for inbox/briefing notes (source:ai-agents-briefing) — blocks 7/9 stuck-note backlog before any fetch


davesnowdon:
## Summary

The repair sweep cannot retry archive download for **inbox/briefing-sourced
notes** (`source:ai-agents-briefing`). `_resolve_archive_download_args`
(`src/influx/repair_hooks.py:685-702`) dispatches only on the `arxiv` and
`rss-*` source tags; any other source hits `_raise_unsupported_source` and the
sweep records a *transient* `unsupported_source` failure — **before the note's
URL is ever read**. Because the failure is transient (uncounted toward the
per-stage cap), these notes retry every sweep forever and never terminalize.

This is the **dominant blocker for the stuck-note backlog**: of the 9 notes
found in the 2026-06-27 audit, **7 are `source:ai-agents-briefing`** and are
blocked here.

> **Correction (supersedes this issue's original framing).** The first version
> said the fix was to "recover the URL from frontmatter." That was misleading:
> these notes are **not** missing their URL, and nothing needs salvaging from the
> content body. See *Data model* below.

## Data model — the URL is already present (structured, doc-level)

A stuck note read back via `lithos_read` (`5a8e8bbc`):

```
metadata.source_url = "https://cycode.com/blog/ai-security-vulnerabilities"   ← present
metadata.source     = null            ← a different field; inbox never sets it
tags: source:ai-agents-briefing, influx:archive-missing, influx:repair-needed,
      text:abstract-only, influx:text-terminal
version: 116                          ← rewritten 116× churning the repair loop
```

- The download URL lives in the **structured doc-level `source_url`** field —
  exactly where the RSS resolver already reads it.
- `_note_source_url` (`repair_hooks.py:574-599`) returns this field (top-level or
  `metadata`). It deliberately does **not** parse a `source_url:` line out of the
  content body — that legacy shape was removed (#176/#187). So there is **no
  "frontmatter recovery"** involved.
- `_resolve_rss_download_args` (`:636-682`) already builds its kwargs from
  `source_url = _note_source_url(note)` (`:647`). A briefing resolver needs the
  same one call.

### What `url=""` in the failure log actually means

`_raise_unsupported_source` (`:450-465`) raises at the **dispatch** step, before
any URL resolution. The hook only populates `url=...` on the download-*failed*
path (`:802`); the unsupported-source path never reads the note's URL. So the
[...]

PR #281, opened 2026-08-02
fix(repair): count unsupported_source toward the archive cap

davesnowdon:
Follow-up to #278, which fixed the terminal-waiver half of the sweep churn but explicitly could not reach this class.

## Problem

A note whose `source:*` tag has no registered reacquirer churns forever. `_reacquirer_for_source` maps only `arxiv` and `rss-*`; anything else returns `None`, so `_raise_unsupported_source` raises `ExtractionError(stage="unsupported_source")` before any network call. That stage classified as **transient**, so `archive_attempts` never advanced, no cap was reached, `influx:repair-needed` was never cleared, and — because the sweep rewrites every visited note to advance retry order (AC-X-8) — the note accrued one version per sweep indefinitely.

Since `updated_at` is also Lithos's recency signal, such notes become permanently the freshest thing in the KB.

Measured in production 2026-07-29 — the two articles that triggered the original investigation:

| Note | Version | Created | `archive_attempts` |
|---|---|---|---|
| `0cff8956` Shared Context Graphs | 108 | 2026-06-23 | **0** |
| `3b1500c5` A2A Protocol v1 2026 | 101 | 2026-06-24 | **0** |

~100 sweeps each, counter never moved. #278 could not help: with no counted failure there is no terminal tag for its waiver to act on.

Worth recording that the archive-policy route also cannot fix this — `[storage.archive_policy.unsupported]` is consulted inside `download_archive`, which is never reached.

## Change

`classify_failure` gains an optional `stage` parameter backed by a `_STAGE_SCOPED_COUNTED_STAGES` table. `unsupported_source` is counted for `stage="archive"` only.

The text-extraction path deliberately keeps classifying it transient — it has its own terminal handling in `_terminate_unsupported_text_source`, which flips `influx:text-terminal` directly rather than through a counter. Making it globally counted would double-handle that path.

With #278's `archive_ok` waiver already merged, affected notes converge in `REPAIR_COUNTED_CAP` (3) sweeps: counter caps → `influx:archive-terminal` → archive condition waived → `influx:repair-needed` cleared → note leaves the sweep set. `influx:archive-missing` is retained, since it stays factually true.

## Tradeoff — this reverses a documented intent

The archive path left the stage transient so a note would recover automatically once a per-source resolver shipped. That auto-recovery is real, but it costs unbounded churn while waiting, and these notes waited 36 days. Retrying cannot help a failure that means "no code path exists for this source" — only shipping a resolver can.

After this, landing a resolver means re-arming affected notes per `docs/operations/runbook.md` §6 — bounded and documented, though currently three manual edits. #280 proposes an operator command for that.

## Test plan

- [x] `make check` green — 2795 passed, ruff + pyright clean
- [x] `TestArchiveScopedCountedStages` — counted for `archive`; transient with no stage, for `tier2`/`tier3`; archive scoping does not widen other transients (`http`, `resolve`, `archive_read`) nor drop globally-counted stages (`parse`, `validate`, `oversize`)
- [x] `TestSweepUnsupportedSourceArchiveConvergence` — reproduces the exact production note shape (`archive-missing` + `abstract-only` + `text-terminal`): below the cap the counter advances and the note stays; at the cap `influx:archive-terminal` flips, `influx:repair-needed` is cleared, `influx:archive-missing` retained
- [x] Both `repair_hooks` docstrings describing the old transient behaviour corrected
- [ ] Post-deploy: `0cff8956` and `3b1500c5` should reach `influx:archive-terminal` after 3 sweeps and stop accruing versions

Closes #279

https://claude.ai/code/session_01QCT69UAC62JFJDX7AcHSAb

### DIFF
--- a/docs/SPECIFICATION.md
+++ b/docs/SPECIFICATION.md
@@ -577,6 +577,20 @@ Tier 2 and Tier 3 hook failures are partitioned into transient (HTTP, transport,
 
 When `tier{N}_attempts` reaches the cap (currently 3), `influx:tier{N}-terminal` is added to the note's tags and that stage is skipped on subsequent sweeps. Transient failures never advance the counter, so flaky network conditions do not burn the cap.
 
+#### Stage-scoped counting: `unsupported_source` on archive
+
+Most discriminators classify the same way everywhere. `unsupported_source` does not — it is **counted for the archive stage only**.
+
+The archive-download hook raises it when no registered adapter owns the note, i.e. `repair_hooks._reacquirer_for_note` returns `None`. That is not a failed attempt but the absence of a code path, so retrying cannot change it — only shipping a resolver can. Left transient, such a note was re-selected and rewritten on every sweep indefinitely (observed in production at v108 after 36 days, with `archive_attempts` still 0), which pinned `updated_at` to "now" and dominated retrieval ranking.
+
+Counted, it converges: `archive_attempts` reaches the cap, `influx:archive-terminal` is applied, the archive clearing condition is waived (§5.3), `influx:repair-needed` is cleared, and the note leaves the sweep set. **`influx:archive-missing` is retained** — it remains factually true that no archive is stored.
+
+Dispatch resolves ownership by source-tag family (`arxiv`; `rss` / `blog` / `rss-<feed>`) **and** by durable identity markers on the note — an `rss-` id, or a `feed-slug:` tag plus a doc-level `source_url`. The identity fallback exists precisely because this failure is now permanent: dispatching on the tag alone would terminalise notes `archive_download_identity` could actually resolve.
+
+Text extraction keeps classifying `unsupported_source` as transient. It has its own terminal path (`repair._terminate_unsupported_text_source` applies `influx:text-terminal` directly rather than through a counter), and counting it globally would double-handle that.
+
+**Operational consequence:** adding a resolver no longer recovers already-terminalised notes automatically. They must be re-armed — see below.
+
 #### Re-arming a terminal stage
 
 **The persisted counter is authoritative, not the tag.** Two different mechanisms read the terminal state, and a re-arm has to satisfy both:
--- a/docs/generated/metrics.json
+++ b/docs/generated/metrics.json
@@ -120,7 +120,7 @@
         "qualname": "influx.promotion_gate.evaluate_promotion_gate"
       }
     ],
-    "total_functions": 652
+    "total_functions": 653
   },
   "domain": {
     "associations": 5,
@@ -420,13 +420,13 @@
       },
       "Repair": {
         "classes": 17,
-        "functions": 79,
+        "functions": 80,
         "largest_module": "influx.repair",
         "largest_module_lines": 1711,
-        "lines": 3839,
+        "lines": 3886,
         "modules": 5,
         "public_symbols": 39,
-        "sloc": 2964
+        "sloc": 3002
       },
       "Schemas": {
         "classes": 4,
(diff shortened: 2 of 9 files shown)


davesnowdon:
## Findings addressed in cfbe643 — with one correction to the diagnosis

`make check` green: **2798 passed**, ruff + pyright clean.

### Finding 1 — valid, fixed. Severity is lower than High, and the production-note hypothesis is wrong.

**The bug is real.** `RssSourceEntry.source_tag` is `Literal["rss", "blog"]`, RSS acquisition persists `blog` verbatim, and **581 production notes carry `source:blog`**. `archive_download_identity` never required an `rss-*` source tag. So counting `unsupported_source` would have terminalised recoverable notes. Good catch — this was the right thing to block on.

Fixed: `_reacquirer_for_source` → `_reacquirer_for_note`, resolving by source-tag family (`arxiv`; `rss` / `blog` / `rss-<feed>`) **and** by durable identity markers (`rss-` id, or `feed-slug` + `source_url`) via the new `_has_rss_identity`. The counted fallback is kept only for notes no adapter can own.

**On severity:** I queried the exact conjunction — **zero** `source:blog` notes currently carry `influx:repair-needed`. The sampled ones are already `influx:archive-terminal` via the archive-policy `unsupported` mode for openai.com. So this was latent with no live instances, not an active High. Still worth fixing before merge, since it's latent in a now-permanent state transition.

**On the cited production notes — this part does not hold.** The review hedges they "may also be false... if they are historical RSS notes carrying feed-slug metadata." They aren't:

```
0cff8956 tags: source:ai-agents-briefing, submitter:ai-agents-briefing,
               influx:archive-missing, influx:repair-needed,
               text:abstract-only, influx:text-terminal, profile:knowledge-systems
```

No `feed-slug:`, and the id is a UUID, not `rss-*`. Same for `3b1500c5`. All 122 `source:ai-agents-briefing` notes carry `submitter:` — the inbox path, which `repair_hooks` already documents as having no resolver (GH #248).

So `_rss_item_id_from_note` returns `None` for them regardless of dispatch. Routing them to `RssSource` swaps `unsupported_source` for `resolve` — **both transient, identical churn**. The counted classification is still what converges them, and the original diagnosis stands.

### Finding 2 — valid, fixed

Four production-wired tests through `make_default_sweep_hooks` with real reacquirers: `source:blog` + feed-slug reaches RssSource (transient network failure, not `unsupported_source`); an unrecognised source tag with feed-slug + source_url also reaches it; and the exact inbox shape behind #279 — UUID id, `submitter:` tag, no feed-slug — is genuinely unsupported and counted. That last one is the test that encodes the diagnosis rather than assuming it.

### Finding 3 — valid, fixed

Now asserts both contracts explicitly: `repair_stage="archive"` → counted, context-free → transient, with a comment covering the manual re-arm requirement.

### Finding 4 — valid, fixed

SPECIFICATION.md §11.1 gains a "Stage-scoped counting" subsection covering archive-scoped counting, retention of `influx:archive-missing`, clearing of `influx:repair-needed`, identity-based dispatch, and the re-arm consequence.

### Nice-to-have taken

`classify_failure(stage=...)` → `repair_stage=`, with a docstring note that `exc.stage` means the failure discriminator. Genuinely confusing as written.

### Answers

1. **Do the cited notes carry feed-slug / rss- id / RSS path?** No to all three — verified above. They are inbox-path notes.
2. **Was `ai-agents-briefing` historically an RSS feed source tag?** No evidence of it. All 122 notes carry `submitter:ai-agents-briefing`, which is the inbox submitter convention, and none carries `feed-slug:`. Their archives, where present, live under `inbox/YYYY/MM/` — e.g. sibling note `23d0dd8d` has `path: inbox/2026/06/f0862f8134.html`.
3. **Is `source:blog` intentionally unsupported?** No — that was the bug. Now supported.
4. **Is manual re-arm accepted policy?** Accepted for now and documented, but it is the weakest part. GitHub #280 proposes an operator command to make it atomic. Automatic reconsideration on registry change would be better still; I have logged it against the follow-up task rather than widening this PR.

### Deferred

"Move ownership detection into the source adapter/registry" — agreed, and it is the right end state; `_has_rss_identity` deepens the duplication it would remove. Tracked as Lithos task `66866df6`, along with distinguishing "unknown source family" from "known source, insufficient identity metadata" in failure stages/telemetry, and the migration mechanism for already-terminalised notes. Not a merge blocker: the layering contract forbids `repair.*` importing `sources.*`, so it needs a `Source.can_reacquire(note)` probe on the injected protocol — a design change rather than a fix.

